PR #7, opened 2019-03-05
Use list instead of deque for replay_buffer

### LINKED ISSUE #5: Use list instead of collections.deque for replay buffer
The current replay buffer at python/social_bot/util/replay_buffer.py uses collections.deque for storage. However, deque can have performance issue when the size is very large. As shown in deque implementation of deque_item https://github.com/python/cpython/blob/3.5/Modules/_collectionsmodule.c#L1123
To access an item, it needs to traverse through all the blocks of size 64, which can be time-consuming when size is large. As I run examples/train_simple_navigation.py, the fps decreases from 45 to 35 and stops decreases after 500,000 steps. It's coincident with the size of replay buffer. So I suspect the longer random accessing time of deque is cause of this problem.

We should use a cycled list to replace deque so that it will not get slower as size gets large.

emailweixu:
fix #5 
With this change, the training does not get slower as more experiences are put in the replay buffer

### DIFF
--- a/python/social_bot/util/replay_buffer.py
+++ b/python/social_bot/util/replay_buffer.py
@@ -18,48 +18,59 @@ class ReplayBuffer(object):
         self._buffer_size = buffer_size
         self._history_length = history_length
         self._future_length = future_length
-        self._buffer = deque(maxlen=buffer_size)
+        self._buffer = [None] * buffer_size
         self._previous_done = 0  # how long ago is the previous done
         # number of experience whose do_not_sample is False
         self._num_valid_experiences = 0
-        self._do_not_sample_flags = deque(maxlen=buffer_size)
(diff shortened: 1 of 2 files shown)


python/social_bot/util/replay_buffer.py
```diff
@@ -18,48 +18,59 @@ def __init__(self, buffer_size, history_length, future_length):
         self._buffer_size = buffer_size
         self._history_length = history_length
         self._future_length = future_length
-        self._buffer = deque(maxlen=buffer_size)
+        self._buffer = [None] * buffer_size
```
bayesian: implement its own cirque class, with [], append() method, would reduce the code change by a lot.

emailweixu: Good idea.

PR #10, opened 2019-03-06
Use list instead of deque for replay_buffer

### LINKED ISSUE #5: Use list instead of collections.deque for replay buffer

bayesian:
Fixes #5 
passed unit tests.

### DIFF
--- a/python/social_bot/util/replay_buffer.py
+++ b/python/social_bot/util/replay_buffer.py
@@ -5,8 +5,49 @@ import random
 from collections import deque
 from .segment_tree import SumSegmentTree, MinSegmentTree
 
+""" Circular array based implementatioon of queue."""
+class CircularBuffer(object):
+    def __init__(self, maxlen, default_value):
+        self._buffer = [default_value] * maxlen
+        self._buffer_size = maxlen
+        self._current_idx = 0
+        self._num_samples = 0
 
-class ReplayBuffer(object):
+    def _idx_to_location(self, idx):
+        idx += self._current_idx
+        if idx >= self._buffer_size:
+            idx -= self._buffer_size
+        return idx
+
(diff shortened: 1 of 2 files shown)


python/social_bot/util/replay_buffer.py
```diff
@@ -158,42 +189,31 @@ def initial_priority(self):
         return self._max_priority
 
     def get_priority(self, idx):
-        idx = self._buffer_idx_to_tree_idx(idx)
+        idx = self.__idx_to_location(idx)
```
emailweixu: __ => _

python/social_bot/util/replay_buffer.py
```diff
@@ -5,8 +5,49 @@
 from collections import deque
 from .segment_tree import SumSegmentTree, MinSegmentTree
 
+""" Circular array based implementatioon of queue."""
+class Cirque(object):
```
emailweixu: CircularBuffer might be a better name?

python/social_bot/util/replay_buffer.py
bayesian: this function not getting called.

PR #13, opened 2019-03-19
migrate from pytorch to tf 2.0

bayesian:
perf is about 41fps. @emailweixu @runjerry

### DIFF
--- a/examples/train_simple_navigation_tf2.py
+++ b/examples/train_simple_navigation_tf2.py
@@ -1,6 +1,8 @@
 # Copyright (c) 2019 Horizon Robotics. All Rights Reserved.
 """
 A demonstration of Q learning for simple_navigation environment
+It is the same as train_simple_navigation.py but implemented in tensorflow 2.0
+
 """
 
 import gym
@@ -15,12 +17,10 @@ import PIL
 from social_bot.util.replay_buffer import PrioritizedReplayBuffer, ReplayBuffer
 from collections import deque, namedtuple
 import time
-
-import torch
-import torch.nn as nn
-import torch.nn.functional as F
-import torch.optim as optim
-
+import tensorflow as tf
+import tensorflow.keras as keras
+import tensorflow.keras.layers as layers
+import tensorflow.keras.models as models
 
 class Options(object):
     """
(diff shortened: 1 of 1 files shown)


examples/train_simple_navigation.py
```diff
@@ -300,10 +304,39 @@ def _update_reward_dist(self):
             i -= 1
         self._replay_buffer.update_priority(indices, priorities)
 
+    #@tf.function
+    #somehow @tf.function will make it 10 times slower
+    def _tf_learn(self, inputs, actions, rewards, next_inputs, dones, reward_dist,
```
emailweixu: It's very strange. I tried putting this into a separate file and using fake date to call it. It's much faster. And using @tf.function is faster than not using it.

emailweixu:
Can the new code train a good policy?
Let's keep the pytorch version as an example.

bayesian:
keep the original pytorch, and create a separated version.

after 9M steps, it got ema_r ~ 0.45 with fps ~ 35

bayesian:
easier to see the changes from: https://github.com/HorizonRobotics/SocialRobot/pull/13/commits/beff3107af77459e9d62766718a729f711e7e663

examples/train_simple_navigation_tf2.py
```diff
@@ -466,107 +484,88 @@ def _make_sample(*exps):
...
         return self.calc_q_values()
 
+    def show_parameter_stats(self):
+        pass
 
 if __name__ == "__main__":
     options = Options()
     os.makedirs(options.model_dir, exist_ok=True)
     logging.basicConfig(level=logging.INFO)
+    logging.getLogger().setLevel(logging.INFO)
```
emailweixu: Adding a comment saying that the logger may have already been created by tensorflow and thus logging.basicConfig() in the previous line won't create a new logger hence `level=logging.INFO` in the previous line does not have any effect. So we need to explicitly set log level here to enable the logging.

examples/train_simple_navigation_tf2.py
```diff
@@ -368,20 +356,18 @@ def learn(self, obs, action, reward, done):
         ema_reward = 0
         if options.ema_reward_alpha > 0:
             ema_reward = self.calc_ema_reward()
+        ema_reward = tf.constant(ema_reward)
```
emailweixu: if options.ema_reward_alpha == 0, the Tensor's dtype will be tf.int32 and it will cause error when it's added to other float Tensor in tf_learn()

PR #15, opened 2019-04-20
Grocery ground pr2

Jialn:
Add the grocery ground environment


Jialn:
If take pose of objects as observation, this environment can be solved by the DDPG example of TF-Agent with 15~20min on my computer. 
The use_image option and teacher part are not tested for now. I'm still working on that.

python/social_bot/envs/grocery_ground.py
```diff
@@ -0,0 +1,232 @@
...
+            port = 0
+        gazebo.initialize(port=port)
+        self._world = gazebo.new_world_from_file(
+            os.path.join(social_bot.get_world_dir(), "grocery_ground.world"))
+        # self._world.info()
+
+        self._agent = self._world.get_agent()
+        self._pr2_joints = list(
+            filter(
+                lambda s: s.find('fl_caster_r_wheel_joint') != -1 or s.find('fr_caster_l_wheel_joint') != -1,
```
bayesian: pr2 has 4 casts: fl,fr,bl,br (front left, front right...), and each has left and right wheel.. it chould have 8 wheel joints to control moving..

python/social_bot/envs/grocery_ground.py
```diff
@@ -0,0 +1,232 @@
...
+        else:
+            sentence = ''
+            controls = action
+        controls = dict(zip(self._pr2_joints, controls))
+        teacher_action = self._teacher.teach(sentence)
+        self._agent.take_action(controls)
+        self._world.step(5)
+        if self._with_language:
+            obs_data = self._get_observation()
+            obs = OrderedDict(data=obs_data, sentence=teacher_action.sentence)
```
emailweixu: How is sentence used by RL algorithm? Right now, it a string which does not match the specification given at line 137

python/social_bot/envs/grocery_ground.py
Jialn: Thanks for the helpful manual. 

To simplify the caster controlling logic, I have make the front rotation joint as fixed in the model file. Now it works as a differential driver like turtle-bot and pioneer. We only need to control 2 wheel_joint in the front and leave other joints free.

python/social_bot/worlds/grocery_ground.world
```diff
@@ -0,0 +1,4624 @@
+<?xml version="1.0" ?>
```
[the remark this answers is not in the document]
Jialn: done

Jialn:
Putting model files (i.e., trash_can) into models/objects folder does not work. Only in model folder can the model be found. Any idea for how to add a sub folder to model database?

python/social_bot/envs/grocery_ground.py
```diff
@@ -0,0 +1,232 @@
...
+from social_bot.teacher import TeacherAction
+from social_bot.envs import simple_navigation
+from social_bot.envs.simple_navigation import GoalTask
+from social_bot.envs.simple_navigation import DiscreteSequence
+import social_bot.pygazebo as gazebo
+
+logger = logging.getLogger(__name__)
+
+
+class GroceryGroundGoalTask(GoalTask):
```
emailweixu: The functionality of this class is almost same as GoalTask. We should update GoalTask and use GoalTask instead. And GoalTask should be moved a separate file as it's not used only by SimpleNavigation

python/social_bot/envs/grocery_ground.py
```diff
@@ -0,0 +1,232 @@
...
+        return obs
+
+    def _get_observation(self):
+        if self._use_image_obs:
+            obs_data = np.array(
+                self._agent.get_camera_observation(
+                    "default::pr2::pr2::head_tilt_link::head_mount_prosilica_link_sensor"
+                ),
+                copy=False)
+            obs_data = PIL.Image.fromarray(rgb2gray(obs_data)).resize(
```
emailweixu: we should use color image. gray image loses information.

python/social_bot/envs/grocery_ground.py
```diff
@@ -0,0 +1,232 @@
...
+        if self._with_language:
+            self.observation_space = gym.spaces.Dict(
+                data=obs_data_space, sentence=DiscreteSequence(128, 24))
+            self._action_space = gym.spaces.Dict(
+                control=control_space, sentence=DiscreteSequence(128, 24))
+        else:
+            self.observation_space = obs_data_space
+            self.action_space = control_space
+
+    def reset(self):
```
emailweixu: Let's create a random map for each episode. We can control how many types of objects are in one map. This should be controlled by Teacher so that teacher can control the task difficulty according to the progress of the agent.

emailweixu: And we should make it easier to switch agent (pr2, pioneer, etc)

python/social_bot/models/pr2_differential_driver/model.config
```diff
@@ -0,0 +1,19 @@
...
+  <name>pr2_differential_driver</name>
+  <version>1.0</version>
+  <sdf version="1.5">model.sdf</sdf>
+
+  <description>
+    Original SDF is from:
+    https://bitbucket.org/osrf/gazebo_models/src/default/pr2/model.sdf
+
+    Modifies: 
+    Fix the front caster rotation joints to simplify the caster controlling logic.Now it works as a differential driver the same as turtle-bot or pioneer.
```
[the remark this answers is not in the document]
Jialn: done

python/social_bot/envs/grocery_ground.py
Jialn: Yes, we should.

python/social_bot/envs/grocery_ground.py
Jialn: done

python/social_bot/envs/simple_navigation.py
```diff
@@ -6,99 +6,20 @@
...
 import logging
-import math
 import numpy as np
 import os
 import random
 
 import social_bot
 from social_bot import teacher
 from social_bot.teacher import TeacherAction
+from social_bot import goal_task
```
emailweixu: This import is not necessary

python/social_bot/envs/grocery_ground.py
```diff
@@ -16,9 +15,8 @@
 import social_bot
 from social_bot import teacher
 from social_bot.teacher import TeacherAction
-from social_bot.envs import simple_navigation
-from social_bot.envs.simple_navigation import GoalTask
-from social_bot.envs.simple_navigation import DiscreteSequence
+from social_bot import goal_task
```
emailweixu: this import is not used.

python/social_bot/envs/grocery_ground.py
```diff
@@ -0,0 +1,247 @@
...
+                   self._goal_random_range / 2,
+                   random.random() * self._goal_random_range -
+                   self._goal_random_range / 2, 0)
+            self._initial_dist = np.linalg.norm(loc - agent_loc)
+            if self._initial_dist > 0.5:
+                break
+        goal.set_pose((loc, (0, 0, 0)))
+
+
+class DiscreteSequence(gym.Space):
```
emailweixu: This is same as the one in simple_navigation. We should not copy it.

python/social_bot/envs/grocery_ground.py
emailweixu: It seems to me that we can add one more argument `goal_random_range" to GoalTask so that we can reuse it instead of implementing a new one.

python/social_bot/goal_task.py
```diff
@@ -0,0 +1,96 @@
+# Copyright (c) 2019 Horizon Robotics. All Rights Reserved.
```
emailweixu: Did you do "pre-commit install" according to https://github.com/HorizonRobotics/SocialRobot/blob/master/CONTRIBUTING.md#workflow

It should automatically insert a full copy right notice like the following instead of one line.
https://github.com/HorizonRobotics/SocialRobot/blob/1b65fda55aaefcd956ee7f1bd60dcbc8c038b5e1/tools/codestyle/copyright.hook#L11-L25

Jialn: Copyright info updated. I did use pre-commit, but it seems that pre-commit won't update if there was already an one-line copyright added by hand.

python/social_bot/envs/grocery_ground.py
Jialn: Move it to teacher class. Both simple_navigation and grocery_gournd import DiscreteSequence from teacher now.

python/social_bot/envs/grocery_ground.py
Jialn: Removed.

python/social_bot/envs/simple_navigation.py
Jialn: Removed.

python/social_bot/goal_task.py
```diff
@@ -27,19 +39,22 @@ def __init__(self,
                  max_steps=500,
                  goal_name="goal",
                  success_distance_thresh=0.5,
-                 fail_distance_thresh=0.5):
+                 fail_distance_thresh=0.5,
+                 goal_random_range=10.0):
```
emailweixu: This change the original behavior, which is goal_random_range=2.0. To keep SimpleNavigation unaffected, you need to change the code there. A better way is to use a default which is same as the original behavior, so the existing code won't need to be changed.

Jialn: Yes, original behavior should not be changed. Fixed now.

emailweixu:
For the last commit "Remove dependency to example of tfagent to avoid importing mujoco", I don't see it's necessary because tf agent won't import mujoco if it does not exist: 

https://github.com/tensorflow/agents/blob/fb8d3309f352c4c5a2788920f20a2453d983c492/tf_agents/environments/suite_mujoco.py#L28

Jialn:
> For the last commit "Remove dependency to example of tfagent to avoid importing mujoco", I don't see it's necessary because tf agent won't import mujoco if it does not exist:
> 
> https://github.com/tensorflow/agents/blob/fb8d3309f352c4c5a2788920f20a2453d983c492/tf_agents/environments/suite_mujoco.py#L28

I have mujoco installed on my ubuntu machine but its license was expired 2 weeks ago. So it reports an error about mujoco license.

Jialn:
Ready to merge now.

python/social_bot/envs/grocery_ground.py
```diff
@@ -0,0 +1,336 @@
...
+
+    def _random_move_objects(self, random_range=10.0):
+        obj_num = len(self._object_types)
+        obj_pos_list = random.sample(self._pos_list, obj_num)
+        for obj_id in range(obj_num):
+            model_name = self._object_types[obj_id]
+            loc = (obj_pos_list[obj_id][0], obj_pos_list[obj_id][1], 0)
+            pose = (np.array(loc), (0, 0, 0))
+            self._world.get_model(model_name).set_pose(pose)
+            self._world.step(10)
```
emailweixu: When I run this environment, sometime I still see an object is located at the same position of the agent (0,0,0), which should not happen.

python/social_bot/envs/grocery_ground.py
```diff
@@ -0,0 +1,336 @@
...
+class GroceryGround(gym.Env):
+    """
+    The goal of this task is to train the agent to navigate to the objects given its
+    name.
+
+    Joints of the agent are controllable by force,
+    the observations are image or the states of the world, including every model's
+    position and rotation.
+
+    Agent will receive a reward provided by the teacher
```
emailweixu: Need to give some more detail about the environment. For example: when does the goal change location, when are the objects rearranged,

examples/train_grocery_ground.py
```diff
@@ -0,0 +1,244 @@
+# Copyright (c) 2019 Horizon Robotics. All Rights Reserved.
+#
+# A simple test for SocialBot-GroceryGround env use TF-Agents
+# The original file is from the PPO example of TF-Agents:
+#
+#     https://github.com/tensorflow/agents
+#
+# Some parameters are modifed suite the environment.
+# Require about 30K episodes/3.6M steps to train the nvaigation task
```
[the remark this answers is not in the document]
Jialn: Done

pygazebo/pygazebo.cc
```diff
@@ -93,6 +93,13 @@ class Model {
                                   std::get<2>(rot));
     model_->SetWorldPose(pose3d);
   }
+  auto GetVelocities() const {
```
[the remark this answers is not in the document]
Jialn: Done

pygazebo/pygazebo.cc
```diff
@@ -396,7 +410,10 @@ PYBIND11_MODULE(pygazebo, m) {
            "Get ((x,y,z), (roll, pitch, yaw)) of the agent")
       .def("set_pose",
            &Model::SetPose,
-           "Set ((x,y,z), (roll, pitch, yaw)) of the agent");
+           "Set ((x,y,z), (roll, pitch, yaw)) of the agent")
+      .def("get_velocities",
+           &Model::GetVelocities,
+           "Get (linear velocity, angular velocity) of the agent");
```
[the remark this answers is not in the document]
Jialn: Done

python/social_bot/envs/grocery_ground.py
```diff
@@ -0,0 +1,336 @@
...
+                 agent_type='pioneer2dx_noplugin',
+                 goal_name='table',
+                 max_steps=160,
+                 port=None):
+        """
+        Args:
+            with_language (bool): the observation will be a dict with an extra sentence
+            use_image_obs (bool): use image as observation, or use pose of the objects
+            agent_type (string): select the agent robot, supporting pr2_differential, 
+                pioneer2dx_noplugin, turtlebot, and create for now
```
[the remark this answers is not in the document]
Jialn: IRobot Create from www.irobot.com

python/social_bot/envs/grocery_ground.py
```diff
@@ -0,0 +1,336 @@
...
+            os.path.join(social_bot.get_world_dir(), "grocery_ground.world"))
+        self._object_types = [
+            'coke_can', 'cube_20k', 'car_wheel', 'plastic_cup', 'beer', 'hammer'
+        ]
+        self._pos_list=list(itertools.product(range(-5, 5), range(-5, 5)))
+        self._pos_list.remove((0,0))
+        self._world.info()
+        self._world.insertModelFile('model://' + agent_type)
+        self._world.step(20)
+        time.sleep(0.1)  # Avoid 'px!=0' error
```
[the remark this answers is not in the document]
Jialn: a multi-thread problem. I will update the comment.

python/social_bot/envs/grocery_ground.py
```diff
@@ -0,0 +1,336 @@
...
+        camera_sensor = {
+            'pr2_differential':
+            'default::pr2_differential::head_tilt_link::head_mount_prosilica_link_sensor',
+            'pioneer2dx_noplugin':
+            ' ',
+            'turtlebot':
+            ' ',
+            'create':
+            ' ',
+        }
```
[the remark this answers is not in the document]
Jialn: OK

python/social_bot/envs/grocery_ground.py
```diff
@@ -0,0 +1,336 @@
...
+            ],
+            'create': ['create::left_wheel', 'create::right_wheel'],
+        }
+        control_force = {
+            'pr2_differential': 20,
+            'pioneer2dx_noplugin': 0.5,
+            'turtlebot': 0.5,
+            'create': 0.5,
+        }
+        # Camera, TODO
```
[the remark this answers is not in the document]
Jialn: There is no camera for pioneer2dx_noplugin  and irobot create. Need to update the sensor name of camera for turtlebot later.

python/social_bot/envs/grocery_ground.py
```diff
@@ -0,0 +1,336 @@
...
+        self._teacher.add_task_group(task_group)
+
+        self._with_language = with_language
+        self._use_image_obs = use_image_obs
+        self._resized_image_size = (84, 84, 3)
+
+        obs = self.reset()
+        if self._use_image_obs:
+            obs_data_space = gym.spaces.Box(
+                low=0, high=1, shape=obs.shape, dtype=np.uint8)
```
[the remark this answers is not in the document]
Jialn: Yes, it should be.

python/social_bot/envs/grocery_ground.py
```diff
@@ -0,0 +1,336 @@
...
+        self._pos_list.remove((0,0))
+        self._world.info()
+        self._world.insertModelFile('model://' + agent_type)
+        self._world.step(20)
+        time.sleep(0.1)  # Avoid 'px!=0' error
+        self._random_insert_objects()
+        self._world.model_list_info()
+
+        # Specify joints and sensors for the robots
+        control_joints = {
```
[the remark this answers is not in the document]
Jialn: Yes, good advice. I will open a PR for this later.

python/social_bot/envs/grocery_ground.py
```diff
@@ -0,0 +1,336 @@
...
+                 with_language=False,
+                 use_image_obs=False,
+                 agent_type='pioneer2dx_noplugin',
+                 goal_name='table',
+                 max_steps=160,
+                 port=None):
+        """
+        Args:
+            with_language (bool): the observation will be a dict with an extra sentence
+            use_image_obs (bool): use image as observation, or use pose of the objects
```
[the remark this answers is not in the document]
Jialn: Done

python/social_bot/envs/grocery_ground.py
Jialn: OK.

python/social_bot/envs/grocery_ground.py
Jialn: The trash can could cause this problem. I can not find the reason. Removed it for now.

python/social_bot/envs/grocery_ground.py
```diff
@@ -95,11 +95,18 @@ class GroceryGround(gym.Env):
     The goal of this task is to train the agent to navigate to the objects given its
     name.
 
+    The envionment support agent type of pr2_differential, pioneer2dx_noplugin, 
+    turtlebot, and irobot create for now. Note that for the models without camera
+    sensor (pioneer, create), you can not use image as observation.
```
emailweixu: pioneer2dx_noplugin has camera

python/social_bot/envs/grocery_ground.py
```diff
@@ -0,0 +1,336 @@
...
+
+
+class GroceryGround(gym.Env):
+    """
+    The goal of this task is to train the agent to navigate to the objects given its
+    name.
+
+    The envionment support agent type of pr2_differential, pioneer2dx_noplugin, 
+    turtlebot, and irobot create for now. Note that for the models without camera
+    sensor (pioneer, create), you can not use image as observation.
```
emailweixu: pioneer2dx_noplugin has camera as simple_navigation uses camera.

python/social_bot/envs/grocery_ground.py
```diff
@@ -0,0 +1,336 @@
...
+                else:
+                    agent_sentence = yield TeacherAction()
+        logger.debug("loc: " + str(loc) + " goal: " + str(goal_loc) +
+                     "dist: " + str(dist))
+        yield TeacherAction(reward=-10.0, sentence="Failed", done=True)
+
+
+class GroceryGround(gym.Env):
+    """
+    The goal of this task is to train the agent to navigate to the objects given its
```
emailweixu: It might mislead reader to think the name of the object is provided to the agent and agent need to understand the object name.

Change to the following:
"...navigate to a fixed type of object. The name of the object is provided in the constructor. In each episode, the location of the goal object is randomly chosen."

Jialn: done

python/social_bot/envs/grocery_ground.py
Jialn: It seems to me that pioneer2dx_noplugin has no camera. Simple_navigation use a separated camera model fixed to pioneer2dx::pioneer2dx::chassis link in pioneer2dx_camera.world.

emailweixu: You are right. We should make a pioneer2dx_with_camara model later.

PR #22, opened 2019-05-13
Gripper

bayesian:
changes: @Jialn @emailweixu @runjerry 

1. control each joint honoring limits (otherwise gripper joint wouldn't open, as its limit is over 10x larger)
2. correct way of implementing contact sensor in pygazebo
3. switch pr2 model.config to base on grocery_ground_pr2 branch, but modify the joint settings related to r_arms.
4. modify friction settings
5. add short tables (shorter than standard one in gazebo)
6. switch from camera-based to internal state based.

With the new env, we could successfully grasp successfully after 4-5M environment steps with tf_agent/ppo

### DIFF
--- a/examples/train_pr2.py
+++ b/examples/train_pr2.py
@@ -8,11 +8,16 @@ from baselines import ppo2
 from baselines import deepq
 from baselines import ddpg
 from baselines.run import main as M
+import tensorflow as tf
 import numpy as np
 
 
 def ppo2():
     M('--alg=ppo2 --env=SocialBot-Pr2Gripper-v0 --num_timesteps=1e6 --network=cnn --save_path=~/models/pr2.model'
+      .split(' '))
+
+def ppo2_mlp():
+    M('--alg=ppo2 --env=SocialBot-Pr2Gripper-v0 --num_timesteps=2e6 --network=lstm --save_path=~/models/pr2_lstm.model --num_env 1 --nminibatches=1 --ent_coef=0.01'
       .split(' '))
 
 
@@ -32,5 +37,7 @@ def eval_ppo2():
 
 
 if __name__ == '__main__':
+    config = tf.ConfigProto()
+    config.gpu_options.allow_growth = True
     logging.basicConfig(level=logging.DEBUG)
-    ppo2_lstm()
+    ppo2_mlp()
--- a/examples/train_pr2.py
+++ b/examples/train_pr2.py
@@ -12,7 +12,7 @@ import tensorflow as tf
 import numpy as np
 
 
-# please refer to alf branch for an successful training setup for PR2 gripper
+# please refer to repo: github.com/HorizonRobotics/alf for a successful training setup for PR2 gripper
 # task using PPO with tf_agents with 60 parallel environments with about
 # 120 frames per sec
 
--- a/examples/train_pr2.py
+++ b/examples/train_pr2.py
@@ -12,6 +12,10 @@ import tensorflow as tf
 import numpy as np
 
 
+# please refer to alf branch for an successful training setup for PR2 gripper
+# task using PPO with tf_agents with 60 parallel environments with about
+# 120 frames per sec
+
 def ppo2():
     M('--alg=ppo2 --env=SocialBot-Pr2Gripper-v0 --num_timesteps=1e6 --network=cnn --save_path=~/models/pr2.model'
       .split(' '))
--- a/examples/train_pr2.py
+++ b/examples/train_pr2.py
@@ -17,12 +17,12 @@ def ppo2():
 
 
 def ppo2_lstm():
-    M('--alg=ppo2 --env=SocialBot-Pr2Gripper-v0 --num_timesteps=2e6 --network=cnn_lstm --save_path=~/models/pr2_lstm.model --num_env 1 --nminibatches=1'
+    M('--alg=ppo2 --env=SocialBot-Pr2Gripper-v0 --num_timesteps=2e6 --network=cnn_lstm --save_path=~/models/pr2_lstm.model --num_env 1 --nminibatches=1 --ent_coef=0.01'
       .split(' '))
 
 
 def eval_ppo2_lstm():
-    M('--alg=ppo2 --env=SocialBot-Pr2Gripper-v0 --num_timesteps=0 --network=cnn_lstm --load_path=~/models/pr2_lstm.model --play --num_env 1 --nminibatches=1'
+    M('--alg=ppo2 --env=SocialBot-Pr2Gripper-v0 --num_timesteps=0 --network=cnn_lstm --load_path=~/models/pr2_lstm.model --play --num_env 1 --nminibatches=1 --ent_coef=0.01'
       .split(' '))
 
 
--- a/pygazebo/pygazebo.cc
+++ b/pygazebo/pygazebo.cc
@@ -186,6 +186,8 @@ class Agent : public Model {
     }
   }
 
+  // given a sensor names, it will return all pairs of link names
+  // which are in collisions detected by the touch sensor at that instant.
   std::set<std::tuple<std::string, std::string>> GetCollisions(
     const std::string& contact_sensor_name) {
 
--- a/pygazebo/pygazebo.cc
+++ b/pygazebo/pygazebo.cc
@@ -175,11 +175,9 @@ class Agent : public Model {
     }
   }
 
-  // TODO: this is not the correct way to expose contacts from gazebo
-  // we could expose api to check whether two links are in contact,
-  // the other way of registering async callback does not seem to fit
-  // in the current pygazebo interface.
-  unsigned int GetCollisionCount(const std::string& contact_sensor_name) {
+  // check contact_sensor whether it touches particular link
+  bool GetContact(const std::string& contact_sensor_name,
+                  const std::string& link_scope_name) {
     auto it = contacts_.find(contact_sensor_name);
 
     if (it == contacts_.end()) {
@@ -198,7 +196,17 @@ class Agent : public Model {
       auto ret = contacts_.insert(std::make_pair(contact_sensor_name, sensor));
       it = ret.first;
     }
-    return it->second->GetCollisionCount();
+
+    auto contacts = it->second->Contacts();
+
+    for (int i = 0; i < contacts.contact_size(); i++) {
+      if (contacts.contact(i).collision1() == link_scope_name ||
+          contacts.contact(i).collision2() == link_scope_name) {
+        return true;
+      }
+    }
+
+    return false;
   }
 
   CameraObservation GetCameraObservation(const std::string& sensor_scope_name) {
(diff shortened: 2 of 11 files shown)


python/social_bot/envs/pr2.py
```diff
@@ -58,128 +81,288 @@ def __init__(self,
...
+
+        if delta_reward > 0:
+            reward += delta_reward
+        else:
+            reward += -0.01     #if no positive reward, penalize it a bit to speed up
+
+        if self._motion_loss > 0.0:
+            v2s = 0.0
+            for joint in self._r_arm_joints:
+                v = self._agent.get_joint_state(joint).get_velocities()[0]
```
Jialn: Maybe we can also simply use actions here. Naturally a human may want to avoid extra efforts (forces of the joints).

python/social_bot/envs/pr2.py
```diff
@@ -58,128 +81,288 @@ def __init__(self,
...
         self._prev_dist = self._get_finger_tip_distance()
-        return self._get_observation()
+        self._prev_gripper_pos = self._get_gripper_pos()
+        self._gripper_reward_dir = 1
+        return obs
 
     def _move_goal(self):
-        loc = (0.6 + 0.3 * (random.random() - 1), 0.4 * (random.random() - 1),
-               0)
+        loc = (0.76 + 0.14 * (random.random() - 1),  0.1 * (random.random() - 1.0),
```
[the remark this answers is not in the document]
bayesian: The PR2 robot is at (0, 0), and this base is pretty big. The table is (0.9, 0), since we only control the right arm, so we put the goal object on the right side (0.76 + 0.14 (random - 0.5)).

python/social_bot/envs/pr2.py
```diff
@@ -58,128 +81,288 @@ def __init__(self,
...
-        self._resized_image_size = (84, 84)
-
-        self._prev_dist = self._get_finger_tip_distance()
-        self._collision_cnt = 0
+        self._resized_image_size = (128, 128) #(84, 84)
+        self._use_internal_states_only = use_internal_states_only
         self._cum_reward = 0.0
+        self._motion_loss = motion_loss
+
+        # a hack to walk around gripper not open initially, might not need now.
```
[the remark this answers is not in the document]
bayesian: yes

python/social_bot/envs/pr2.py
```diff
@@ -58,128 +81,288 @@ def __init__(self,
...
+            obs = np.concatenate((
+                np.array(joint_positions), np.array(joint_velocities), np.array(loc),
+                np.array(loc_a), np.array(l_finger_tip_loc), np.array(l_finger_tip_loc_a),
+                np.array(r_finger_tip_loc), np.array(r_finger_tip_loc_a),
+                np.array([l_dist, r_dist]),
+                np.array([self._l_touch, self._r_touch]).astype(np.float32)), 0)
+        else:
+            obs = np.array(
+                self._agent.get_camera_observation(
+                    "default::pr2::pr2::head_tilt_link::head_mount_prosilica_link_sensor"
```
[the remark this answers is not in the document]
bayesian: agree, currently only the internal states observation works, we shall investigate into stereo camera later.

python/social_bot/envs/pr2.py
bayesian: yes, ||f||_2 could be better here, but we did not use it yet.

python/social_bot/envs/pr2.py
```diff
@@ -58,128 +81,288 @@ def __init__(self,
...
+
+        # final actions used to control each joints are:
+        # joint_limit_{i} *  \pi (state)_{i}
+        actions = scaled_actions
+
+        controls = dict(zip(self._r_arm_joints, actions))
 
         self._agent.take_action(controls)
         self._world.step(100)
+
```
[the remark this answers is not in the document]
bayesian: just follow the convention as other environments, also I don't think we need such frequency controls here.. 10-20 steps have two problems: frequency too high 50-100 hz, and also episode and simulation too slow.. 10x longer

python/social_bot/envs/pr2.py
```diff
@@ -38,17 +38,40 @@ class Pr2Gripper(gym.Env):
...
+                 motion_loss=0.0000,
+                 use_internal_states_only=True,
                  port=None):
         """
         Args:
             goal_name (string): name of the object to lift off ground
             max_steps (int): episode will end when the agent exceeds the number of steps.
             reward_shaping (boolean): whether it adds distance based reward shaping.
+            motion_loss (float): if not zero, it will add -motion_loss * || V_{all_joints} ||^2 in
+                 the reward when episode endds.
```
emailweixu: endds => ends

pygazebo/pygazebo.cc
```diff
@@ -440,8 +462,8 @@ PYBIND11_MODULE(pygazebo, m) {
            py::arg("sensor_scope_name"))
       .def("get_joint_state", &Agent::GetJointState, py::arg("joint_name"))
       .def("get_link_pose", &Agent::GetLinkPose, py::arg("link_name"))
-      .def("get_collision_count",
-           &Agent::GetCollisionCount,
+      .def("get_collisions",
+           &Agent::GetCollisions,
```
emailweixu: Add a help message to describe the function in .def()

python/social_bot/envs/pr2.py
```diff
@@ -58,128 +81,288 @@ def __init__(self,
...
 
         self._all_joints = self._agent.get_joint_names()
 
-        self._world.info()
+        # to avoid different parallel simulation has the same randomness
+        random.seed(port)
+
+        #self._world.info()
+
+        # from https://github.com/PR2/pr2_mechanism/blob/melodic-devel/pr2_mechanism_model/pr2.urdf
```
emailweixu: comment what a passive joint is

python/social_bot/envs/pr2.py
```diff
@@ -58,128 +81,288 @@ def __init__(self,
...
+        gripper_pos = self._get_gripper_pos()
 
         done = self._steps_in_this_episode >= self._max_steps
 
-        reward = -0.01 if (not done) else (1.2 - dist)
+        reward = 0 if (not done) else 1.0 - np.tanh(5.0 * dist)
+
+        dist_reward = self._prev_dist - dist
+
+        pos_reward = self._gripper_reward_dir * (gripper_pos - self._prev_gripper_pos)
```
emailweixu: Give a description about this reward

python/social_bot/envs/pr2.py
```diff
@@ -58,128 +81,288 @@ def __init__(self,
...
+        r_finger_tip_loc = np.array(self._r_finger_pose[0])
 
-        dist = np.linalg.norm(np.array(finger_tip_loc) - np.array(goal_loc))
+        finger_tip_center = 0.5 * (l_finger_tip_loc + r_finger_tip_loc)
+        dist = np.linalg.norm(np.array(finger_tip_center) - np.array(goal_loc))
         return dist
 
-    def step(self, action):
-        controls = dict(zip(self._r_arm_joints, action))
+    def _get_gripper_pos(self):
```
emailweixu: comment that this is the angle

python/social_bot/envs/pr2.py
```diff
@@ -58,128 +81,288 @@ def __init__(self,
...
+        state = self._agent.get_joint_state("pr2::pr2::r_gripper_joint")
+        return state.get_positions()[0]
+
+    def step(self, actions):
+        def trunc_scale(a, limit):
```
emailweixu: min(max(a, -1), 1) * limit

python/social_bot/envs/pr2.py
```diff
@@ -38,17 +38,40 @@ class Pr2Gripper(gym.Env):
     currently, the agent only learns to use arm/hand to play bat to hit the goal
```
emailweixu: And describe how the reward is given.

python/social_bot/envs/pr2.py
```diff
@@ -58,128 +81,288 @@ def __init__(self,
...
+                                           "beer::link::box_collision")
+        self._goal_pose = self._goal.get_pose()
+        self._l_finger_pose = self._agent.get_link_pose("pr2::pr2::r_gripper_l_finger_tip_link")
+        self._r_finger_pose = self._agent.get_link_pose("pr2::pr2::r_gripper_r_finger_tip_link")
+
+        if self._use_internal_states_only:
+            joint_states = [self._agent.get_joint_state(joint) for joint in self._r_arm_joints]
+            joint_positions = [s.get_positions()[0] for s in joint_states]
+            joint_velocities = [s.get_velocities()[0] for s in joint_states]
+            loc, loc_a = self._goal_pose
```
emailweixu: should consider using logical camera to get the relative pose of the objects in the view

examples/train_pr2.py
```diff
@@ -32,5 +41,7 @@ def eval_ppo2():
 
 
 if __name__ == '__main__':
+    config = tf.ConfigProto()
+    config.gpu_options.allow_growth = True
     logging.basicConfig(level=logging.DEBUG)
-    ppo2_lstm()
+    ppo2_mlp()
```
Jialn: Is that from baseline? I remember you said that tf-agent ppo was used and many parameters are tuned.

python/social_bot/envs/pr2.py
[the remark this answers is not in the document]
Jialn: the max_step_size option in world file should be higher to keep time scale the same. 
i.e., max_step_size 0.001 -> 0.005,  if 100substeps -> 20 substeps. Control frequency won't change here, but the simulator will be 5x faster.

examples/train_pr2.py
bayesian: it is in alf depot, will send out PR shortly.

python/social_bot/envs/pr2.py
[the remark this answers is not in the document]
emailweixu: Cannot understand your response

pygazebo/pygazebo.cc
emailweixu: Can you also add a help message for this function?

python/social_bot/envs/pr2.py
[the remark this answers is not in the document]
bayesian: I will try that when train with camera based observation

pygazebo/pygazebo.cc
bayesian: done

pygazebo/pygazebo.cc
```diff
@@ -166,11 +177,11 @@ class Agent : public Model {
     }
   }
 
-  // TODO: this is not the correct way to expose contacts from gazebo
-  // we could expose api to check whether two links are in contact,
-  // the other way of registering async callback does not seem to fit
-  // in the current pygazebo interface.
-  unsigned int GetCollisionCount(const std::string& contact_sensor_name) {
+  std::set<std::tuple<std::string, std::string>> GetCollisions(
```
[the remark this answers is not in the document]
bayesian: done

PR #27, opened 2019-05-31
Make simple_navigation support different channel order

emailweixu:
Also fix a bug of pioneer2dx_camera.world of incorrect link name

### DIFF
--- a/python/social_bot/envs/simple_navigation.py
+++ b/python/social_bot/envs/simple_navigation.py
@@ -11,16 +11,18 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-"""
-A simple enviroment for navigation.
-"""
+"""A simple enviroment for navigation."""
+
 from collections import OrderedDict
-import gym
-import gym.spaces
 import logging
 import numpy as np
 import os
 import random
+
+import gym
+import gym.spaces
+import gin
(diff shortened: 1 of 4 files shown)


python/social_bot/envs/simple_navigation.py
```diff
@@ -11,17 +11,19 @@
...
-import gym
-import gym.spaces
 import logging
 import numpy as np
 import os
 import random
 
+import gym
+import gym.spaces
+import gin
```
hnyu: Should we also update the dependency for 'gin'?

emailweixu: tf_agents already depends on gin

emailweixu: You mean for setup.py?

hnyu: I mean if someone wants to use SocialRobot independently, then the 'gin' dependency is not met.

setup.py
```diff
@@ -11,8 +11,7 @@
 setup(
     name='social_bot',
     version='0.0.1',
-    install_requires=['gym', 'numpy',
-                      'matplotlib'],  # And any other dependencies foo needs
+    install_requires=['gin-config', 'gym', 'numpy', 'matplotlib'],
```
hnyu: The README has manual installation for these libs. Maybe change it to running setup.py? (also include 'psutil')

emailweixu: Changed

PR #32, opened 2019-06-04
Support take_action with PID controller

Jialn:
Support take_action with PID controller, and slightly adjust grocery_ground env


pygazebo/pygazebo.cc
```diff
@@ -264,6 +265,24 @@ class Agent : public Model {
...
+  bool TakeActionPDController(const std::map<std::string, double>& ctrl_target) {
+    // Create a PID controller
+    // The parameters suits for revolute/continuous joint, from error of radian to force
+    auto pid = gazebo::common::PID(0.02, 0, 0.01);
+    pid.SetCmdMax(10.0);
+    pid.SetCmdMin(-10.0);
+    auto controller = model_->GetJointController();
+    for (const auto& name2target : ctrl_target) {
+      controller->SetVelocityPID(name2target.first, pid);
+      bool ret = controller->SetVelocityTarget(name2target.first, name2target.second);
```
emailweixu: We should set PID separately in a different function and expose that to python SDK because the PID parameters might be different for different joints.

pygazebo/pygazebo.cc
```diff
@@ -264,6 +265,24 @@ class Agent : public Model {
     }
   }
 
+  bool TakeActionPDController(const std::map<std::string, double>& ctrl_target) {
```
emailweixu: we can merge this function with TakeAction(), adding another arguments: velocities

pygazebo/pygazebo.cc
Jialn: Sure, we should. Will update it later.

pygazebo/pygazebo.cc
Jialn: Do we need postion controlled joints? For some sinario like rotating camera. we might want a position control.

emailweixu: We can add it either now or later.

pygazebo/pygazebo.cc
```diff
@@ -472,10 +490,24 @@ PYBIND11_MODULE(pygazebo, m) {
...
+           &Agent::SetPIDController,
+           "Set PID parameters for a joint if PID controleer are used"
+           "joint_name is the name for the joint"
+           "p, i, and d are the parameters for the controller"
+           "max_force is the limit of PID contorller output",
+           py::arg("joint_name"),
+           py::arg("p") = 0.02,
+           py::arg("i") = 0.00,
+           py::arg("d") = 0.01,
+           py::arg("max_force") = 2.0)
```
emailweixu: keep the default max_force same as the default value in the function definition

pygazebo/pygazebo.cc
```diff
@@ -252,18 +253,37 @@ class Agent : public Model {
             reinterpret_cast<const uint8_t*>(sensor->ImageData())));
   }
 
-  bool TakeAction(const std::map<std::string, double>& forces) {
+  bool TakeAction(const std::map<std::string, double>& controls,
```
emailweixu: As you suggested, we may add position control later. So we may use position target to control some joint, while using velocity to control some other joints. If we want to allow that possibility, we should separate different arguments for these different control mode: force, velocity, and position

pygazebo/pygazebo.cc
```diff
@@ -252,18 +254,59 @@ class Agent : public Model {
...
   }
 
-  bool TakeAction(const std::map<std::string, double>& forces) {
+  bool TakeAction(const std::map<std::string, double>& controls) {
     auto controller = model_->GetJointController();
-    for (const auto& name2force : forces) {
-      bool ret = controller->SetForce(name2force.first, name2force.second);
+    for (const auto& name2control : controls) {
+      bool ret;
+      int control_type = joints_control_type_[name2control.first];
```
emailweixu: need to default the control_type to control_type_force_ because the user may not always call SetPIDController()

Jialn: It's safe here. Value in std::map<std::string, int> is 0 (Forces) by default, and it will be created automaticly at the line control_type = joints_control_type_[jointname] if that jointname is not existing. Setting defaut explicitly when agent is creating needs to repeat the operation GetJointNames() since we don't know the joint list, so we can save some initializing codes.

Std::map overload operator "[ ]", it will find the key. but if the key is not existing, then create one with default value, int() here so it's 0. (So we cannot use joint_control_type.find here).

emailweixu: You are right. I confused it with python map.

PR #34, opened 2019-06-06
Implement render() for the enviroments

emailweixu:


emailweixu:
Most changes in grocery_ground.py are caused by auto-formatting, except the part related to EnvBase

python/social_bot/envs/cartpole.py
```diff
@@ -111,14 +114,15 @@ def reset(self):
...
+    env._world.info()
     for _ in range(100):
-        print("reset")
         env.reset()
-        env._world.info()
         total_rewards = 0
         while True:
             obs, reward, done, info = env.step((random.random() - 0.5) * 0.5)
             total_rewards += reward
+            env.render()
```
Jialn: When run out of 100 episodes before terminating, it throws an error on my PC(Ubuntu 16.04, Gazebo 9.6.0):
```
python: /usr/include/boost/thread/pthread/recursive_mutex.hpp:113: void boost::recursive_mutex::lock(): Assertion `!pthread_mutex_lock(&m)' failed.
Aborted (core dumped)
```

python/social_bot/envs/base.py
```diff
@@ -0,0 +1,49 @@
...
+# limitations under the License.
+
+import os
+
+import gym
+
+import social_bot.pygazebo as gazebo
+
+
+class EnvBase(gym.Env):
```
Jialn: Is "base.py" dedicated to implement gazebo rendering? If so, maybe we could change the file name to something like "render_base.py" or "gazeboenv_base.py" .

Jialn:
> Most changes in grocery_ground.py are caused by auto-formatting, except the part related to EnvBase

Seems only 7 line changes in grocery_ground.py and all of them are related to EnvBase. Do you mean pr2.py which has 144 changes?

emailweixu:
Right, I made a mistake. It's not grocery_ground.py, it's pr2 which has lot of changes due to auto-format.

python/social_bot/envs/base.py
emailweixu: Changed.

PR #40, opened 2019-06-23
Gro2 env: support random goal from teacher task

Jialn:
Add a random goal support, goal's name (string) was encoded by hash and remapped to range (-1, +1)

Some encoded codes for example:
goalname:hammer, encode to: -0.2531430738150149
goalname:bookshelf, encode to: 0.43231639941069755
goalname:cube_20k, encode to: 0.7890891626763448
goalname:car_wheel, encode to: 0.25891776869694905
goalname:table, encode to: 0.33379767746182765
goalname:beer, encode to: -0.5148801851630069
goalname:plastic_cup, encode to: 0.1974983606824552

### DIFF
--- a/python/social_bot/envs/grocery_ground.py
+++ b/python/social_bot/envs/grocery_ground.py
@@ -21,6 +21,7 @@ import numpy as np
 import random
 import PIL
 import itertools
+import hashlib
 
 import gym
 from gym import spaces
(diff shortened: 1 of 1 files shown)


emailweixu:
I think a more general approach would be better: the Task yield the utterance, then teacher change string of word sequence to integer sequence with fixed length, which can be used for models.

Encoding using a random number is not good for two reasons: 1. it cannot be extend to future scenario of sentences, 2. using a fixed random number to represent the goal may make the learning more difficult.

Jialn:
> I think a more general approach would be better: the Task yield the utterance, then teacher change string of word sequence to integer sequence with fixed length, which can be used for models.
> 
> Encoding using a random number is not good for two reasons: 1. it cannot be extend to future scenario of sentences, 2. using a fixed random number to represent the goal may make the learning more difficult.

Ok, I will close this PR and re-implement this with "with_sentence=True", as a minimal example of future scenario of full sentences.

Jialn:
> I think a more general approach would be better: the Task yield the utterance, then teacher change string of word sequence to integer sequence with fixed length, which can be used for models.
> 
> Encoding using a random number is not good for two reasons: 1. it cannot be extend to future scenario of sentences, 2. using a fixed random number to represent the goal may make the learning more difficult.


Which way is better to encode the word sequence to integer sequence? Shall we use one-hot encoding?

emailweixu:
Just a sequence of integer is fine. The encoding is handled by the model.

PR #42, opened 2019-06-25
Gro2env: random goal from teachers task

Jialn:
Added a helper function to class DiscreteSequence, to convert a sentence to integer sequence by subword segmentation based on [BPEmb(Byte-Pair Encoding Embedding)](https://nlp.h-its.org/bpemb/). 

Segmentation & Encoding Examples(1000 vocab size):
```
plastic_cup  ['▁t', 'able'][181 191  36 976 924 138   0   0   0   0   0   0   0   0   0   0   0   0   0   0]
car_wheel ['▁car', '_', 'w', 'he', 'el'] [403 976 931   5  48   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0]
coke_can ['▁c', 'ok', 'e', '_', 'c', 'an'][ 14 270 913 976 924  16   0   0   0   0   0   0   0   0   0   0   0   0   0   0]
table   ['▁t', 'able'] [  3 383   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0  0   0]
please go to table  ['▁p', 'le', 'ase', '▁go', '▁to', '▁t', 'able']
```

if using a 10000 vocal size:
```
"plastic_cup"  ['▁plastic', '_', 'c', 'up']
"coke_can" ['▁c', 'oke', '_', 'can']
"car wheel" ['▁car', '▁wheel']
"table" ['▁table']
"please go to table" ['▁ple', 'ase', '▁go', '▁to', '▁table']
```

### DIFF
--- a/python/social_bot/envs/grocery_ground.py
+++ b/python/social_bot/envs/grocery_ground.py
@@ -48,7 +48,7 @@ class GroceryGroundGoalTask(teacher_tasks.GoalTask):
         """
         Args:
             reward_shaping (bool): if ture, use shaped reward accroding to distance rather than -1 and 1
-	        random_goal (bool): if ture, teacher will randomly select goal from the object list each episode
+            random_goal (bool): if ture, teacher will randomly select goal from the object list each episode
         """
         super(GroceryGroundGoalTask, self).__init__(**kwargs)
         self._reward_shaping = reward_shaping
@@ -113,7 +113,7 @@ class GroceryGroundGoalTask(teacher_tasks.GoalTask):
                     reward=0.0
                 agent_sentence = yield TeacherAction(
                     reward=reward,
-                    sentence="Please go to "+self._goal_name,
+                    sentence=self._goal_name,  # "Please go to "+self._goal_name,
                     done=False)
         yield TeacherAction(reward=0.0, sentence="Failed", done=True)
 
@@ -143,7 +143,7 @@ class GroceryGround(GazeboEnvBase):
     def __init__(self,
                  with_language=False,
                  use_image_obs=False,
-	             random_goal=False,
+                 random_goal=False,
                  use_pid=False,
                  agent_type='pioneer2dx_noplugin',
                  max_steps=160,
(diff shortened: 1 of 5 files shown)


python/social_bot/envs/grocery_ground.py
```diff
@@ -81,12 +108,13 @@ def run(self, agent, world):
...
-                        reward=-dist / self._random_range,
-                        sentence="Failed",
-                        done=False)
+                    reward = -dist / self._random_range
                 else:
-                    agent_sentence = yield TeacherAction()
+                    reward=0.0
+                agent_sentence = yield TeacherAction(
+                    reward=reward,
+                    sentence="Please go to "+self._goal_name,
```
emailweixu: Right now it's ok to only use the goal name in the sentence (i.e. without "Please go to"). We'll do full sentences later.

python/social_bot/envs/grocery_ground.py
```diff
@@ -141,10 +170,20 @@ def __init__(self,
         super(GroceryGround, self).__init__(port=port)
         self._world = gazebo.new_world_from_file(
             os.path.join(social_bot.get_world_dir(), "grocery_ground.world"))
-        self._object_types = [
-            'coke_can', 'table', 'bookshelf', 'cube_20k', 'car_wheel',
-            'plastic_cup', 'beer', 'hammer'
-        ]
+
+        self._teacher = teacher.Teacher(False)
```
emailweixu: Use a keyword for "False" to make it easier to read the code.

python/social_bot/envs/grocery_ground.py
```diff
@@ -115,9 +143,10 @@ class GroceryGround(GazeboEnvBase):
     def __init__(self,
                  with_language=False,
                  use_image_obs=False,
+	             random_goal=False,
+                 use_pid=False,
```
emailweixu: comment of these arguments

python/social_bot/envs/grocery_ground.py
```diff
@@ -195,20 +223,20 @@ def __init__(self,
...
             shape=[len(self._agent_joints)],
             dtype=np.float32)
 
         if self._with_language:
             self.observation_space = gym.spaces.Dict(
-                data=obs_data_space, sentence=DiscreteSequence(128, 24))
+                data=obs_data_space, sentence=DiscreteSequence(256, 24))
             self.action_space = gym.spaces.Dict(
-                control=control_space, sentence=DiscreteSequence(128, 24))
+                control=self._control_space, sentence=DiscreteSequence(256, 24))
```
emailweixu: We still need to convert this to integer sequence somewhere. If it's easy to do on alf side (i.e. tensorflow has related ops), we can do it on alf side. Otherwise, we need to convert sentence to integer sequence.

python/social_bot/envs/grocery_ground.py
Jialn: OK

python/social_bot/envs/grocery_ground.py
Jialn: Done. It's task_groups_exclusive=False .

python/social_bot/envs/grocery_ground.py
Jialn: Done

Jialn: Done

python/social_bot/envs/grocery_ground.py
Jialn: I added a helper function to class DiscreteSequence, to convert a sentence to integer sequence by subword segmentation based on BPEmb.

emailweixu:
Using sub-word id is a good idea. But I think we should still have a baseline of using word id, which is the standard practice for NLP. I understand that having to specify the vocabulary is annoying. We can have a property for each Task for the vocab used by that task, and let the teacher merge the vocab from every tasks.

PR #57, opened 2019-07-16
ALF training examples

Jialn:
Changes:

- Add ALF training examples
- Code cleaning for the task class in grocery_ground.py
- Removed unused pr2 baselines and grocery tfagents-ppo examples


emailweixu:
for alf training, can the robot achieve the task goal after training?

examples/grocery_alf_ppo.gin
```diff
@@ -0,0 +1,48 @@
+import alf.algorithms.ppo_algorithm
+import alf.algorithms.ppo_loss
+import social_bot
+
+# training & playiing
```
emailweixu: playiing => playing

examples/icub_alf_ppo.gin
```diff
@@ -0,0 +1,49 @@
+# training & playiing with Agent Learning Framework (Alf)
```
emailweixu: playiing => playing

python/social_bot/envs/grocery_ground.py
```diff
@@ -286,6 +238,9 @@ def reset(self):
         self._world.reset()
         self._teacher.reset(self._agent, self._world)
         self._random_move_objects()
+        if self._random_goal:
+            random_id = random.randrange(len(self._object_list))
+            self._teacher_task.set_goal_name(self._object_list[random_id])
```
emailweixu: goal selection should be done by the teacher.

Jialn:
> for alf training, can the robot achieve the task goal after training?

Yes, most of the time.  Sometimes fails because agent is blocked by other models in the environment 
zvjs6-aj32h

examples/grocery_alf_ppo.gin
```diff
@@ -27,8 +27,8 @@ ActorDistributionNetwork.activation_fn=@tf.nn.tanh
 ActorDistributionNetwork.continuous_projection_net=@NormalProjectionNetwork
 NormalProjectionNetwork.init_means_output_factor=1e-10
 NormalProjectionNetwork.std_bias_initializer_value=0.0
-NormalProjectionNetwork.scale_distribution=True
-estimated_entropy.assume_reparametrization=True
+# NormalProjectionNetwork.scale_distribution=True
+# estimated_entropy.assume_reparametrization=True
```
emailweixu: Is this used or not? If not, it should be removed. And ALF master now should support this now.

python/social_bot/envs/grocery_ground.py
```diff
@@ -62,6 +62,12 @@ def __init__(self, reward_shaping=False, random_goal=False, **kwargs):
         ]
         self.task_vocab = self.task_vocab + self._objects_in_world + self._objects_to_insert
 
+    def run(self, agent, world):
+        if self._random_goal:
+            random_id = random.randrange(len(self._objects_to_insert))
+            self.set_goal_name(self._objects_to_insert[random_id])
+        yield super(GroceryGroundGoalTask, self).run(agent, world)
```
emailweixu: Did you test this? I think the right way to do this is: yield from super(...).run(...). See https://docs.python.org/3/whatsnew/3.3.html#pep-380

emailweixu: And we can use the simpler new syntax: super().run(agent, world)

Jialn: My mistake. Should use "yield from" here.

examples/grocery_alf_ppo.gin
Jialn: Removed. I was meant to use this, but it seems current version is good enough as a simple example.

PR #59, opened 2019-07-19
WIP: add icub alf sac example

Jialn:


Jialn:
(Test result, gin file is @28af537)
Agent could keep not falling up to 200 steps in 800-900K env steps, pretty faster. But reward is lower, just keep standing, does not walking forward. And the performance goes down periodically. 
As to the speed, Alf SAC 4.5M env steps cost ~ 2days, about 3x slower than  tf-agent SAC. 

Alf SAC, gin file @28af537
image

tf-agent SAC, the ICubWalkDefault one
image

Jialn:
performance also goes down periodically in case of PPO.

image

emailweixu:
My training using [ppo_icubwalk.gin](https://github.com/HorizonRobotics/alf/blob/master/alf/examples/ppo_icubwalk.gin) in alf repo is stable.

My experience so far is that state_dependent_std=True makes the training unstable. I am still investigating the reason.

PR #61, opened 2019-07-23
Some minor changes

Jialn:
Some minor changes like using action_space.sample() instead of np.random(action.shape), use np.clip instead of min(max()), etc.


Jialn:
Also added rendering mode "rgb_array". Since gzclient might not be always opening, and we can not directly access user camera of gzclient in gzserver, this mode is done by inserting a camera to the world. 

The default view pose is like below. User can set view point by setting pose of "self._rendering_camera" in "env.step()" .
image

Jialn:
updated the camera sdf string:
image

python/social_bot/envs/gazebo_base.py
```diff
@@ -48,8 +51,45 @@ def render(self, mode='human'):
...
                 self._rendering_process = Popen(['gzclient'])
             return
+        if mode == 'rgb_array':
+            if self._rendering_camera is None:
+                render_camera_sdf = """
+                <?xml version='1.0'?>
+                <sdf version ='1.4'>
+                <model name ='render_camera'>
+                    <static>1</static>
+                    <pose>-3 0 3 0 0.4 0</pose>
```
emailweixu: A fixed pose probability does not fit for all different environment. We should allow different environment have different pose.

emailweixu:
Does alf.trainer.on_policy_trainer.play() generate right video from this?

emailweixu:
It seems that you need to add metadata like this so the video can be captured:
https://github.com/bulletphysics/bullet3/blob/0aaae872451a69d0c93b0c8ed818667de4ad5653/examples/pybullet/gym/pybullet_envs/env_bases.py#L16

python/social_bot/envs/gazebo_base.py
emailweixu: Can you choose right pose for each of current environments?

Jialn: Sure.

Jialn:
> It seems that you need to add metadata like this so the video can be captured:
> https://github.com/bulletphysics/bullet3/blob/0aaae872451a69d0c93b0c8ed818667de4ad5653/examples/pybullet/gym/pybullet_envs/env_bases.py#L16

OK， I'll add it.

Jialn:
Because the rendered image has no grid, and the yellow texture in texture_ground_plane draws too much attention, I changed the yellow texture to grey texture with 1m x 1m grid:

image

Jialn:
the default view point, this also suits for icubwalk and simple navi:
Screenshot from 2019-07-24 16-00-27
PR2 has been set to a closer camera pos
image

python/social_bot/envs/pr2.py
```diff
@@ -382,6 +382,8 @@ def run(self):
         while True:
             actions = self.action_space.sample()
             obs, r, done, _ = self.step(actions * self._gripper_reward_dir)
+            #self.render('rgb_array')human
```
emailweixu: remove

emailweixu:
Do the model trained with the previous ground plane texture work with the new ground texture?

emailweixu:
And since the ground texture is changed. Perhaps we should update the videos in  README.md

PR #68, opened 2019-08-21
Multi task_group support for grocery ground env

Jialn:
multi task_group support for grocery ground with an example second task of icub auxiliary task

### DIFF
--- a/python/social_bot/envs/grocery_ground.py
+++ b/python/social_bot/envs/grocery_ground.py
@@ -170,7 +170,7 @@ class ICubStandingTask(GroceryGroundTaskBase):
         """
         super().__init__()
         self.reward_weight = reward_weight
-        self.task_vocab = ['icub', 'walk']
+        self.task_vocab = ['icub']
 
     def _get_contacts_to_ground(self, contacts_sensor):
         contacts = self._agent.get_collisions(contacts_sensor)
@@ -386,7 +386,8 @@ class GroceryGround(GazeboEnvBase):
                  agent_type='pioneer2dx_noplugin',
                  port=None,
                  resized_image_size=(64, 64),
-                 data_format='channels_last'):
+                 data_format='channels_last',
+                 vocab_sequence_length=20):
         """
         Args:
             with_language (bool): The observation will be a dict with an extra sentence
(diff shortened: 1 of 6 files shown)


python/social_bot/teacher.py
```diff
@@ -147,6 +150,9 @@ def add_task_group(self, task_group, weight=1):
         self._task_groups.append(task_group)
         self._weights.append(weight)
 
+    def get_task_group(self):
```
emailweixu: get_task_groups

python/social_bot/envs/grocery_ground.py
```diff
@@ -515,7 +577,11 @@ def _get_camera_observation(self):
         return img
 
     def _get_low_dim_full_states(self):
-        task_specific_ob = self._teacher_task.task_specific_observation()
+        task_specific_ob = np.array([])
+        for task_group in self._teacher.get_task_group():
+            for gro_task in task_group.get_tasks():
+                task_specific_ob = np.append(task_specific_ob,
+                    gro_task.task_specific_observation())
```
emailweixu: should put line 580-584 to Teacher

Jialn: Yes, I understand. But not all tasks need specified observation, only tasks inherited from "GroceryGroundTaskBase" have this function defined for now. Should we keep it this way or add abstract "task_specific_observation()" to Task base class?

python/social_bot/envs/grocery_ground.py
```diff
@@ -170,6 +171,60 @@ def task_specific_observation(self):
         return np.array(goal.get_pose()[0]).flatten()
 
 
+class ICubStandingTask(GroceryGroundTaskBase):
+    """
+    A simple task spicified for iCub, to keep the agent from falling down.
+    """
+    
+    def __init__(self):
+        super()
```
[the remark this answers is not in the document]
Jialn: yes, added `.__init__()`

python/social_bot/teacher.py
Jialn: changed.

python/social_bot/envs/grocery_ground.py
emailweixu: Adding to Task class is better.

python/social_bot/envs/grocery_ground.py
```diff
@@ -262,7 +322,7 @@ def run(self, agent, world):
                         # within 45 degrees of the agent direction
                         hitted_ball = True
                 agent_sentence = yield TeacherAction(
-                    reward=-dist / self._random_range)
+                    reward=-dist / self._random_range + 0.5)
```
Jialn: Add an offset of 0.5, so that the expected return is not negative. Negative reward might cause the agent prefer to get self-killed ('done' and reset) ASAP to get rid of more negative reward.

python/social_bot/teacher.py
```diff
@@ -79,6 +81,16 @@ def run(self, agent, world):
...
+    @abstractmethod
+    def task_specific_observation(self):
+        """
+        Args:
+            None
+        Returns:
+            np.array of the extra observations should be added into the
+            observation besides self states, for the non-image case
+        """
+        pass
```
emailweixu: Since  subclass does have to implement this, we can return [] here and remove `@abstractmethod` decorator and comment that subclass can override this

python/social_bot/envs/grocery_ground.py
```diff
@@ -191,35 +195,47 @@ def run(self, agent, world):
...
+                joint_state = self._agent.get_joint_state(joint_name)
+                joint_pos.append(joint_state.get_positions())
+            joint_pos = np.array(joint_pos).flatten()
+            movement_cost = np.sum(np.abs(joint_pos)) / joint_pos.shape[0]
+            reward = 2.0 * alive_reward - 0.5 * movement_cost
             agent_sentence = yield TeacherAction(reward=reward, done=done)
-
-    def task_specific_observation(self):
+    
+    @staticmethod
```
Jialn: Because we need to use the average pos of some key point for agent icub to calculate the distance to goal, static-method "get_icub_extra_obs()" was added to  class ICubStandingTask. So that we can access additional infomation of icub in other tasks.

Jialn:
Changes for the last commit  1c18bfb:
Added action cost, action cost was added to main class, we can set it through gin configuration.
Using some of the tricks from roboschool humanoid flagrun. 
Fix typo.

For the training example, the agent still can not reach the ball after 3M env steps. But can walk a few step to get close to the ball, and if we fix the position of ball the agent can walk to it. 

We can check in this PR in now. I'll open another PR for some reward shaping for steering and update the training example .

PR #72, opened 2019-08-29
Add angle-to-target regulation for ICubAuxiliaryTask

Jialn:
Changes:
update comments;
move gin parameters from commands to gin file;
add angle-to-target regulation to grocery icub aux task, to face towards the target;


python/social_bot/envs/simple_navigation.py
```diff
@@ -157,9 +157,9 @@ def step(self, action):
             sentence = ''
             controls = action
         controls = dict(zip(self._joint_names, controls))
-        teacher_action = self._teacher.teach(sentence)
```
Jialn: I'm not very sure here. It seems "teacher.teach()" should be after the call of "world.step()". As the teacher_action.reward should be the calculated from the observation after action being taken.

python/social_bot/envs/grocery_ground.py
```diff
@@ -608,9 +640,9 @@ def step(self, action):
             action_ctrl = action
         controls = np.clip(action_ctrl, -1.0, 1.0) * self._agent_control_range
         controls = dict(zip(self._agent_joints, controls))
-        teacher_action = self._teacher.teach(sentence)
         self._agent.take_action(controls)
         self._world.step(self._sub_steps)
+        teacher_action = self._teacher.teach(sentence)
```
Jialn: Similar to above comment. I'm not very sure here.

python/social_bot/envs/grocery_ground.py
```diff
@@ -252,19 +261,40 @@ def task_specific_observation(self):
...
         """
-        return self.get_icub_extra_obs(self._agent)
+        icub_extra_obs =  self.get_icub_extra_obs(self._agent)
+        agent_pos = icub_extra_obs[:3]
+        agent_speed = (agent_pos - self._pre_agent_pos) / self._step_time
+        self._pre_agent_pos = agent_pos
+        yaw = self._agent.get_pose()[1][2]
+        target_pos, _ = self._target.get_pose()
+        yaw = yaw % (2 * np.pi) - np.pi  # model icub has a built-in 180 degree rotation
+        walk_target_theta = np.arctan2(target_pos[1] - agent_pos[1], target_pos[0] - agent_pos[0])
```
Jialn: make use of some tricks from [roboschool forward walker](https://github.com/openai/roboschool/blob/42d1c1c14f57b58f156072b6b5f9b761ecaf3018/roboschool/gym_forward_walker.py#L60)

python/social_bot/envs/grocery_ground.py
```diff
@@ -484,11 +568,10 @@ def __init__(self,
                 max_steps=200,
                 success_distance_thresh=0.5,
                 fail_distance_thresh=3.0,
-                random_goal=with_language,
-                random_range=10.0)
```
Jialn: isame as default range, and will override gin config, so removed.

python/social_bot/envs/grocery_ground.py
```diff
@@ -191,20 +207,49 @@ def run(self, agent, world):
...
+        x = self._agent_init_pos[0] + random.random() * self._random_range
+        y = self._agent_init_pos[1] + random.random() * self._random_range
+        agent_pos = np.array([x, y, 0.6])
+        # a trick from roboschool, to encourage optimal straight walk
+        orient = 0
+        if random.randint(0, 2) == 0:
+            orient = self._get_angle_to_target(agent_pos, 'iCub::root_link',
+                                               np.pi)
+        pose = (agent_pos, np.array([0, 0, orient]))
+        # one problem here is we can not control the sequence of task run(), if
```
Jialn: one problem here is we can not control the sequence of task run(), if this aux task is latter than main task, the target might be at initial pose and has not been randomly moved yet.

python/social_bot/envs/grocery_ground.py
```diff
@@ -191,20 +207,49 @@ def run(self, agent, world):
...
+        self._pre_agent_pos = self.get_icub_extra_obs(agent)[:3]
         agent_sentence = yield
         done = False
+        # set icub initial pose
+        x = self._agent_init_pos[0] + random.random() * self._random_range
+        y = self._agent_init_pos[1] + random.random() * self._random_range
+        agent_pos = np.array([x, y, 0.6])
+        # a trick from roboschool, to encourage optimal straight walk
+        orient = 0
+        if random.randint(0, 2) == 0:
```
Jialn: This trick comes form [roboschool_humanoid_flagrun](https://github.com/openai/roboschool/blob/42d1c1c14f57b58f156072b6b5f9b761ecaf3018/roboschool/gym_humanoid_flagrun.py#L22)
It is helpful to learn to steer and walk straightly, and speed up the learning progress a lot. The agent can walk to the ball only after about 2 hours training.

python/social_bot/envs/grocery_ground.py
emailweixu: Right, it should be after step()

python/social_bot/envs/grocery_ground.py
```diff
@@ -191,20 +207,49 @@ def run(self, agent, world):
...
+            orient = self._get_angle_to_target(agent_pos, 'iCub::root_link',
+                                               np.pi)
+        pose = (agent_pos, np.array([0, 0, orient]))
+        # one problem here is we can not control the sequence of task run(), if 
+        # this aux task is latter than main task, the target might be at initial
```
emailweixu: latter => later

python/social_bot/envs/grocery_ground.py
emailweixu: Right now, teacher go through the task groups in the same order as they are added to teacher. So it should not be a problem.

python/social_bot/envs/grocery_ground.py
```diff
@@ -191,20 +207,49 @@ def run(self, agent, world):
...
+        # pose and has not been randomly moved yet. Should find a way
+        # to fix the call order of run() for the tasks
+        agent.set_pose(pose)
```
emailweixu: It doesn't seem right to "force" the agent change pose in this way. At least we need to let the agent to change its pose after some time. If steering the agent is necessary, we can reduce the probability of doing it gradually to 0.

python/social_bot/envs/grocery_ground.py
```diff
@@ -191,20 +207,49 @@ def run(self, agent, world):
...
             for joint_name in self._joints:
                 joint_state = self._agent.get_joint_state(joint_name)
                 joint_pos.append(joint_state.get_positions())
             joint_pos = np.array(joint_pos).flatten()
             movement_cost = np.sum(np.abs(joint_pos)) / joint_pos.shape[0]
-            reward = 3.0 * alive_reward - 0.5 * movement_cost
+            # orientation cost, the agent should face towards the target
+            # only orientation of root link is not enough here
+            agent_pos = self.get_icub_extra_obs(agent)[:3]
+            head_angel = self._get_angle_to_target(agent_pos, 'iCub::head')
```
emailweixu: angel => angle, here and other places

python/social_bot/envs/grocery_ground.py
Jialn: This trick comes form [roboschool_humanoid_flagrun](https://github.com/openai/roboschool/blob/42d1c1c14f57b58f156072b6b5f9b761ecaf3018/roboschool/gym_humanoid_flagrun.py#L22)
A compromise to algorithm, not so elegant. I'll remove it.

python/social_bot/envs/grocery_ground.py
Jialn: fixed

python/social_bot/envs/grocery_ground.py
emailweixu: The problem of placing it after world.step() is that if teacher changes the state of the world, the agent may not see it immediately. And of course, placing it before world.step() has its own problem.

Jialn: Maybe we can separate the original teacher.teach() to 2 functions like teach() and give_reward(), and place teach() before world.step()?

Jialn: Oh, seems the separation won't help. Agent get observation from last step 
```
step(agent_action, agent_sentence):
    agent.take_action(agent_action)
    teacher_sentence = teacher.teach(agent_sentence) # give sentence and might do some changes to world
    world.step()
    reward = teacher.give_reward() # give reward
    agent_observation = teacher_sentence + observation_from_world
    return agent_observation, reward
```

```
step(agent_action, agent_sentence):
    agent.take_action(agent_action)
    world.step()
    reward, teacher_sentence = teacher.teach(agent_sentence) # give reward, sentence and might do some changes to world
    agent_observation = teacher_sentence + observation_from_world
    return agent_observation, reward
```

le-horizon: Just thinking out loud, shouldn't step() first let agent observe and take action, and world.step(), and then teacher teach?  Reward is part of world.step()?

If agent action happens before observation, then the first action will be kind of random?

Probably doesn't affect performance that much though.

emailweixu: Teacher will do first teach() in reset(), then agent will get the observation from reset and generate actions to call env.step()

Jialn: So we can keep it this way, put teach() before step().

emailweixu: One possible way is to put teacher.teach() between two world.step() so that teacher can get teach based on the immediate result from the actions of the learner and give reward and change the world state if necessary.
```python
    self._agent.take_action(controls)
    self._world.step(self._sub_steps/2, skip_rendering)
    teacher_action = self._teacher.teach(sentence)
    self._world.step(self._sub_steps/2)
```
But I am not absolutely sure about whether this change won't complicate the logic in teacher or not.
```

Comment (le-horizon) [reply python/social_bot/envs/grocery_ground.py]: I think JT's second option is more natural?
If teacher wants to change world state (e.g. switch to a different task), the teacher can do a world.reset() inside teach()?  If the teacher is continuing the same task, then maybe the teacher need to see the effect of agent's action before giving the instruction, and the teacher may also change the env, so it's natural that the agent will delay seeing the effect of the teacher changing the world by 1 world.step().

Does that make sense?


```
step(agent_action, agent_sentence):
    agent.take_action(agent_action)
    world.step()
    reward, teacher_sentence = teacher.teach(agent_sentence) # give reward, sentence and might do some changes to world
    agent_observation = teacher_sentence + observation_from_world
    return agent_observation, reward
```

Jialn: I think we can keep it this simple way for now, If necessary we can change to the half step formula (or add option for it) anytime in the future.

PR #74, opened 2019-09-03
Pr use MultiDiscrete instead of DiscreteSequence so gym _spec_from_gym_space won't complain

le-horizon:
Make groceryground more configurable.  Update input type from DiscreteSequence to MultiDiscrete for language task (because tf_agents only supports using their predefined types during _spec_from_gym_space).

### DIFF
--- a/python/social_bot/__init__.py
+++ b/python/social_bot/__init__.py
@@ -46,6 +46,12 @@ register(
 )
 
 register(
+    id='SocialBot-GroceryGroundImage-v1',
+    entry_point='social_bot.envs:GroceryGroundImage',
+    max_episode_steps=1e8,
+)
+
+register(
     id='SocialBot-GroceryGroundLanguage-v0',
     entry_point='social_bot.envs:GroceryGroundLanguage',
     max_episode_steps=200,
--- a/python/social_bot/__init__.py
+++ b/python/social_bot/__init__.py
@@ -64,6 +64,12 @@ register(
 )
 
 register(
+    id='SocialBot-GroceryGroundImageLanguage-v1',
+    entry_point='social_bot.envs:GroceryGroundImageLanguage',
+    max_episode_steps=1e8,
+)
+
+register(
     id='SocialBot-GroceryGroundImageSelfStatesLanguage-v0',
     entry_point='social_bot.envs:GroceryGroundImageSelfStatesLanguage',
     max_episode_steps=200,
--- a/python/social_bot/__init__.py
+++ b/python/social_bot/__init__.py
@@ -46,12 +46,6 @@ register(
 )
 
 register(
-    id='SocialBot-GroceryGroundImage-v1',
-    entry_point='social_bot.envs:GroceryGroundImage',
-    max_episode_steps=1e8,
-)
-
-register(
     id='SocialBot-GroceryGroundLanguage-v0',
     entry_point='social_bot.envs:GroceryGroundLanguage',
     max_episode_steps=200,
@@ -61,12 +55,6 @@ register(
     id='SocialBot-GroceryGroundImageLanguage-v0',
     entry_point='social_bot.envs:GroceryGroundImageLanguage',
     max_episode_steps=200,
-)
-
-register(
-    id='SocialBot-GroceryGroundImageLanguage-v1',
-    entry_point='social_bot.envs:GroceryGroundImageLanguage',
-    max_episode_steps=1e8,
 )
 
 register(
--- a/python/social_bot/envs/grocery_ground.py
+++ b/python/social_bot/envs/grocery_ground.py
@@ -70,7 +70,7 @@ class GroceryGroundGoalTask(GroceryGroundTaskBase, GoalTask):
 
     def __init__(self,
                  max_steps=500,
-                 goal_name="goal",
+                 goal_name="ball",
                  success_distance_thresh=0.5,
                  fail_distance_thresh=0.5,
                  random_range=2.0,
@@ -86,6 +86,8 @@ class GroceryGroundGoalTask(GroceryGroundTaskBase, GoalTask):
             random_range (float): the goal's random position range
             random_goal (bool): if ture, teacher will randomly select goal from the object list each episode
         """
+        if goal_name is None:
+            goal_name = 'ball'
         GoalTask.__init__(
             self,
             max_steps=max_steps,
@@ -95,7 +97,7 @@ class GroceryGroundGoalTask(GroceryGroundTaskBase, GoalTask):
             random_range=random_range)
         GroceryGroundTaskBase.__init__(self)
         self._random_goal = random_goal
-        self._goal_name = 'ball'
+        self._goal_name = goal_name
         self._objects_in_world = [
             'placing_table', 'plastic_cup_on_table', 'coke_can_on_table',
             'hammer_on_table', 'cafe_table', 'ball'
@@ -104,6 +106,8 @@ class GroceryGroundGoalTask(GroceryGroundTaskBase, GoalTask):
             'coke_can', 'table', 'bookshelf', 'car_wheel', 'plastic_cup',
             'beer', 'hammer'
         ]
+        logging.info("goal_name %s, random_goal %d, fail_distance_thresh %f",
+            self._goal_name, self._random_goal, fail_distance_thresh)
         self._pos_list = list(itertools.product(range(-5, 5), range(-5, 5)))
         self._pos_list.remove((0, 0))
         self.reward_weight = reward_weight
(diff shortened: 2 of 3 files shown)


emailweixu:
Conflicts need to be resolved. The title of this PR does not seem right.

le-horizon:
Oops, forgot to pull before requesting review..  Resolved conflict, and tested image+language training.

Could you take another look?

Thanks,
Le

python/social_bot/__init__.py
```diff
@@ -57,6 +63,12 @@
     max_episode_steps=200,
 )
 
+register(
+    id='SocialBot-GroceryGroundImageLanguage-v1',
+    entry_point='social_bot.envs:GroceryGroundImageLanguage',
+    max_episode_steps=1e8,
```
emailweixu: There is no need to register another one if the difference with v0 is only max_episode_steps, max_episode_steps can be configured by suite_gym.load.max_episode_steps=1000

python/social_bot/envs/grocery_ground.py
```diff
@@ -434,6 +435,9 @@ def __init__(self,
                  image_with_internal_states=False,
                  task_name='goal',
                  agent_type='pioneer2dx_noplugin',
+                 random_goal=None,
+                 fail_distance_thresh=3,
+                 max_steps=200,
```
emailweixu: all of these can be configured through GroceryGroundGoalTask

python/social_bot/envs/grocery_ground.py
```diff
@@ -611,6 +625,8 @@ def step(self, action):
             if type(sentence) != str:
                 sentence = self._teacher.sequence_to_sentence(sentence)
             action_ctrl = action['control']
+            #if not self._train:
+            #    tf.compat.v1.logging.info("Teacher sentence: " + sentence)
```
emailweixu: remove

python/social_bot/__init__.py
le-horizon: Good to know!

python/social_bot/envs/grocery_ground.py
le-horizon: Oh, somehow missed that.

le-horizon: Actually these are fixed around line 497 when initializing the GroceryGroundGoalTask.  This change makes it configurable.

le-horizon:
updated code again.  Some of the changes to make gin configurable is still needed..

python/social_bot/envs/grocery_ground.py
```diff
@@ -86,6 +86,8 @@ def __init__(self,
             random_range (float): the goal's random position range
             random_goal (bool): if ture, teacher will randomly select goal from the object list each episode
         """
+        if goal_name is None:
```
Jialn: it seems this can be removed

python/social_bot/envs/grocery_ground.py
Jialn: I also think it is better to configure through gin of GroceryGroundGoalTask.  GroceryGround is meant for many task, some of them does not need random_goal or fail_distance_thresh.  We can removed the parameters when initializing the task at line 497.

Jialn: As to max_steps,  it should also be removed here. We have the gym parameter max_episode_steps so don't bother with this parameter. And most time we should use max_episode_steps, so that the final return of the last step can be correctly handled:
If max_episode_steps is reached and the env is reset, the potential discounted future reward will be take into account when update the value network. Most task should be like this. 
If max_steps is reached with 'done' and the env is reset, the algorithm don't know env reaches time limits and treat this the same as done. All potential discounted reward in the future will be considered as zero. For some task like goal task, that also could work. but using max_episode_steps is better.
If max_steps is reached without 'done' and the env is reset, the algorithm part don't know env is reset, the initial state of next episode will be considered as the next state of the last step. This should never happen.

You can configure the max_episode_steps like this: suite_socialbot.load.max_episode_steps=300

python/social_bot/envs/grocery_ground.py
```diff
@@ -480,11 +492,13 @@ def __init__(self,
 
         self._teacher = teacher.Teacher(task_groups_exclusive=False)
         if task_name is None or task_name == 'goal':
+            if random_goal is None:
```
Jialn: this can be removed

python/social_bot/envs/grocery_ground.py
le-horizon: Good observation Jiangtao.  I put it here just in case the user passes in `goal_name = None` in the gin file, in which case, a default value is still provided here.

le-horizon: Maybe it's better to alert the user if that happens?

python/social_bot/envs/grocery_ground.py
```diff
@@ -480,11 +492,13 @@ def __init__(self,
...
+                random_goal = with_language
             main_task = GroceryGroundGoalTask(
-                max_steps=200,
+                max_steps=max_steps,
```
Jialn: We can remove 498-502, and make the default parameter of GroceryGroundGoalTask the same as the initialization here. 
We can config GroceryGroundGoalTask through gin when needed.

python/social_bot/envs/grocery_ground.py
Jialn: It's ok just remove it. No one would set goal_name=None in gin file...

python/social_bot/envs/grocery_ground.py
```diff
@@ -70,7 +70,7 @@ class GroceryGroundGoalTask(GroceryGroundTaskBase, GoalTask):
 
     def __init__(self,
```
Jialn: The default "__init__()" parameter could be modified to be the same as initializing code, to keep the default behavior the same after removed the initialing code.

python/social_bot/envs/grocery_ground.py
le-horizon: Done.

python/social_bot/envs/grocery_ground.py
le-horizon: Good point!  Done.

python/social_bot/envs/grocery_ground.py
le-horizon: Done.

le-horizon:
Good points Jiangtao and good to know about the complexity involved in stopping episodes.
Done as you suggested.  Could you take another look?

Thanks,
Le

python/social_bot/envs/grocery_ground.py
```diff
@@ -500,9 +497,13 @@ def __init__(self,
             icub_standing_task = ICubStandingTask()
             icub_aux_task_group.add_task(icub_standing_task)
             self._teacher.add_task_group(icub_aux_task_group)
+        self._teacher._build_vocab_from_tasks()
```
Jialn: This might need to be removed. _build_vocab_from_tasks is done within teacher.py automatically when "teacher.sentence_to_sequence()" is called

le-horizon: I need the vocab_size before sentence_to_sequence() is called.  Unless I missed something, this call is needed?

Jialn: Sorry. I have not noticed that : D

Jialn:
Is there any chance to import DiscreteSequence to alf, and make it support language task? Seems multi-discrete action is made for game controllers or keyboards. I think it would be better to have a gym-space designed for language, and is convenient to extend to other encoding/embedding in the future, like sub-words or something else.

emailweixu:
MultiDiscrete seems ok. If we change to use pretrained continuous word embedding, we can use Box space.

> Is there any chance to import DiscreteSequence to alf, and make it support language task? Seems multi-discrete action is made for game controllers or keyboards. I think it would be better to have a gym-space designed for language, and is convenient to extend to other encoding/embedding in the future, like sub-words or something else.

python/social_bot/envs/grocery_ground.py
le-horizon: Good suggestions.  All done.

le-horizon:
> Is there any chance to import DiscreteSequence to alf, and make it support language task? Seems multi-discrete action is made for game controllers or keyboards. I think it would be better to have a gym-space designed for language, and is convenient to extend to other encoding/embedding in the future, like sub-words or something else.

Unfortunately it's not an alf requirement.  It's tf_agents requiring only supported classes: https://github.com/tensorflow/agents/blob/master/tf_agents/environments/gym_wrapper.py#L107

Jialn:
OK. Seems we have no choice   :-)
Shall we remove DiscreteSequence and use multi-discrete if it is not useful?

le-horizon:
Hi Wei, still need your approval.

For DiscreteSequence, I think we at least need to make a comment saying currently tf_agents requires us to use MultiDiscrete.  Maybe it will be supported in the future?

PR #75, opened 2019-09-04
allow logging teacher and task actions via gin

le-horizon:
Jiangtao and Wei, could you take a look at this small change?  I've debated on whether doing it just for playing mode, not training mode, but that seem to require a lot more code and less flexible.

Gin configuration is more flexible and also easier to code.

Thanks,
Le

### DIFF
--- a/python/social_bot/teacher.py
+++ b/python/social_bot/teacher.py
@@ -223,7 +223,7 @@ class Teacher(object):
     _task_groups_exclusive = True
     vocab_size = 0
 
-    def __init__(self, task_groups_exclusive=True, log_teacher_actions=False):
+    def __init__(self, task_groups_exclusive=True):
         """Create a Teacher instance.
 
         Args:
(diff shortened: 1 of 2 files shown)


Jialn:
Maybe we can remove the optional arg and just log it. We can control the output of logging by setting loglevel at alf side. This is also very convenient, like: `logging.set_verbosity(logging.DEBUG)`
It's also easier to set to log for playing only, together with some other logging info which we might also want to see.

le-horizon:
Interesting, is there a central place where we can use gin to configure set_verbosity?  I'm fine with either approach.  Let me know.

Jialn:
Now we set the verbosity at alf/bin/play or alf/bin/train, actually the first line to run of the python file.
By default it is logging.INFO. Change it to logging.DEBUG to turn on more log for debug.

le-horizon:
Sounds good Jiangtao.  I will make a separate change to use FLAG to control verbosity in command line in alf.

Would you guys prefer the teacher's sentence to be logged at level debug?  Info might be too much during training.

Thanks,
Le

le-horizon:
updated code, with accompany change from alf: https://github.com/HorizonRobotics/alf/pull/165

PR #77, opened 2019-09-06
reset speed to 0 after resetting target

le-horizon:
Currently the ball is sometimes rolling after reset.  It seems this change makes the ball less likely to roll after reset.
Maybe rolling can still happen if ball is initialized to be overlapping with another object in the grocery ground.

After pulling this change, folks will need to do the following to get grocery_ground working:

```
cd SocialRobot_REPO_ROOT
cd build
cmake ..
make -j```

if ```cmake ..``` complains about cannot find configuration file provided by "gazebo", run
```sudo apt install libgazebo9-dev``` where the version number 9 comes from ```gazebo --version```

### DIFF
--- a/pygazebo/pygazebo.cc
+++ b/pygazebo/pygazebo.cc
@@ -135,6 +135,9 @@ class Model {
         std::make_tuple(lin_ver.X(), lin_ver.Y(), lin_ver.Z()),
         std::make_tuple(ang_ver.X(), ang_ver.Y(), ang_ver.Z()));
   }
+  void Reset() {
+    model_->ResetPhysicsStates();
+  }
 };
 
 class Agent : public Model {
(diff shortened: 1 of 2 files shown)


Jialn:
Yes, we only set pose of goal each time. The speed of the ball is kept. 
There is a success_distance_thresh which will end the episode before the agent knocking the ball. But somehow sometimes the agent knock the ball and cause it get a speed. 
Actually there is a bug which will be fixed in PR #72 : teacher.teach() should be called after world.step() for both simplenavi and grocery env. So there is 1 step offset of reward calculation and done checking now. This will not be a problem after the bug is fixed.

le-horizon:
Glad you found the bug of teach before step, Jiangtao!
Did you say you tested target reset already?  If so, I'll just delete this PR.

Jialn:
> Glad you found the bug of teach before step, Jiangtao!
> Did you say you tested target reset already? If so, I'll just delete this PR.

Let's check in this pr. It's good to have a model reset function in pygazebo.

PR #82, opened 2019-10-07
Pr allow curriculum learning in grocery ground goal task

le-horizon:
Hey Jiangtao, Wei and Haonan, this change allows curriculum teaching of goal task by increasing random_range every time agent's success rate is above certain threshold (e.g. 0.9).

It also allows mixing in some percentage (e.g. 20%) of full random_range goals, in which case, those instances are not used to compute the agent's success rate.

Could you take a look?

Thanks,
Le

### DIFF
--- a/python/social_bot/envs/grocery_ground.py
+++ b/python/social_bot/envs/grocery_ground.py
@@ -75,6 +75,7 @@ class GroceryGroundGoalTask(GroceryGroundTaskBase, GoalTask):
                  fail_distance_thresh=3,
                  random_range=10.0,
                  random_goal=False,
+                 use_curriculum_training=False,
                  start_range=0,
                  increase_range_by_percent=50.,
                  reward_thresh_to_increase_range=0.4,
@@ -90,6 +91,7 @@ class GroceryGroundGoalTask(GroceryGroundTaskBase, GoalTask):
                 it's considered a failure and is givne reward -1
             random_range (float): the goal's random position range
             random_goal (bool): if ture, teacher will randomly select goal from the object list each episode
+            use_curriculum_training (bool): when true, use curriculum in goal task training
             start_range (float): for curriculum learning, the starting random_range to set the goal
                 Enables curriculum learning if start_range > 1.2 * success_distance_thresh.
                 NOTE: Because curriculum learning is implemented using teacher in the environment,
@@ -110,6 +112,7 @@ class GroceryGroundGoalTask(GroceryGroundTaskBase, GoalTask):
             success_distance_thresh=success_distance_thresh,
             fail_distance_thresh=fail_distance_thresh,
             random_range=random_range,
+            use_curriculum_training=use_curriculum_training,
             start_range=start_range,
             increase_range_by_percent=increase_range_by_percent,
             reward_thresh_to_increase_range=reward_thresh_to_increase_range,
--- a/python/social_bot/envs/grocery_ground.py
+++ b/python/social_bot/envs/grocery_ground.py
@@ -602,7 +602,7 @@ class GroceryGround(GazeboEnvBase):
         elif task_name == 'kickball':
             main_task = GroceryGroundKickBallTask(step_time=step_time)
         else:
-            logging.debug("upsupported task name: " + task_name)
+            logging.debug("unsupported task name: " + task_name)
 
         main_task_group = TaskGroup()
         main_task_group.add_task(main_task)
--- a/python/social_bot/envs/grocery_ground.py
+++ b/python/social_bot/envs/grocery_ground.py
@@ -128,6 +128,9 @@ class GroceryGroundGoalTask(GroceryGroundTaskBase, GoalTask):
             'coke_can', 'table', 'bookshelf', 'car_wheel', 'plastic_cup',
             'beer', 'hammer'
         ]
+        self._goals = self._objects_to_insert
+        if self._random_goal:
+            self._goals = self._goal_name.split(',')
         logging.info("goal_name %s, random_goal %d, fail_distance_thresh %f,",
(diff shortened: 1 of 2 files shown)


python/social_bot/envs/grocery_ground.py
```diff
@@ -85,6 +90,17 @@ def __init__(self,
                 it's considered a failure and is givne reward -1
             random_range (float): the goal's random position range
             random_goal (bool): if ture, teacher will randomly select goal from the object list each episode
+            start_range (float): for curriculum learning, the starting random_range to set the goal
+                Enables curriculum learning if start_range > 1.2 * success_distance_thresh.
```
Jialn: Maybe we can add a bool flag like "use_curriculum_learning" to enable/disable, and make the default to false. This way is more clear.

python/social_bot/envs/grocery_ground.py
```diff
@@ -85,6 +90,17 @@ def __init__(self,
...
+                NOTE: Because curriculum learning is implemented using teacher in the environment,
+                currently teacher status are not stored in model checkpoints.  Resuming is not supported.
```
Jialn: As curriculum range is increased according to parameter reward_thresh_to_increase_range automatically, is it supposed to be kind of supporting resuming?

python/social_bot/envs/grocery_ground.py
```diff
@@ -509,6 +533,9 @@ def __init__(self,
                  agent_type='pioneer2dx_noplugin',
                  world_time_precision=None,
                  step_time=0.1,
+                 random_goal=None,
+                 fail_distance_thresh=3,
+                 max_steps=200,
```
Jialn: Seems these 3 are not used.

python/social_bot/teacher_tasks.py
```diff
@@ -50,15 +56,57 @@ def __init__(self,
...
+            self._q.popleft()
+        self._q.append(value)
+        if (value > 0 and len(self._q) == self._max_reward_q_length and
+            sum(self._q) >= self._max_reward_q_length *
+                self._reward_thresh_to_increase_range):
+            self._random_range *= 1. + self._increase_range_by_percent
+            if self._random_range > self._orig_random_range:
+                self._random_range = self._orig_random_range
+            logging.info("Raising random_range to %f", self._random_range)
+            self._q.clear()
```
Jialn: I suggest polyak average can be used here, which could make the code logic simpler and less computational, and can have similar effect. 

```
alpha = 0.001
self.polyak_reward = value * alpha + self.polyak_reward * (1 - alpha)
if self.polyak_reward > reward_thresh_to_increase_range:
    self._random_range += self._random_range * self._increase_range_by_percent
    self.polyak_reward = 0
```
alpha has similar effect to 'max_reward_q_length' here.

python/social_bot/teacher_tasks.py
```diff
@@ -50,15 +56,57 @@ def __init__(self,
...
         self.task_vocab = ['hello', 'goal', 'well', 'done', 'failed', 'to']
 
+    def should_use_curriculum_training(self):
+        return self._start_range >= self._success_distance_thresh * 1.2
+
+    def _push_reward_queue(self, value):
+        if (not self.should_use_curriculum_training() or
+            self._is_full_range_in_curriculum):
+            return
+        while len(self._q) >= self._max_reward_q_length:
```
emailweixu: deque has an argument maxlen. You don't need to pop it if maxlen is provided.

python/social_bot/teacher_tasks.py
```diff
@@ -50,15 +56,57 @@ def __init__(self,
...
+            self._q.popleft()
+        self._q.append(value)
+        if (value > 0 and len(self._q) == self._max_reward_q_length and
```
emailweixu: I think "len(self._q) == self._max_reward_q_length" can be removed. It's unlikely to exceed the reward_thresh without the queue being full.

python/social_bot/teacher_tasks.py
emailweixu: I think using queue if Ok, it's easier to reason about the effect, plus the code isn't that much complex if maxlen is used.

python/social_bot/teacher_tasks.py
le-horizon: Nice!

python/social_bot/teacher_tasks.py
le-horizon: Sounds good.

python/social_bot/teacher_tasks.py
le-horizon: Interesting, Jiangtao, I'll keep this in mind for the future.  For this one, I'll just use success rate?

python/social_bot/envs/grocery_ground.py
le-horizon: Done

le-horizon:
Thanks for the comments guys.  Please take another look.

python/social_bot/teacher_tasks.py
le-horizon: Oh, actually after curriculum advances, we clear the deque, and it's very likely the agent can get a few episodes successfully (because agent can already pass earlier level in curriculum), and pass the next level of the curriculum by accident.

python/social_bot/teacher_tasks.py
Jialn: Yes, of course, just a simple suggestion :-)

python/social_bot/envs/grocery_ground.py
Jialn: It seems that random_goal，fail_distance_thresh and max_steps are not used

PR #84, opened 2019-10-18
Replace irobot create with kuka youbot

Jialn:
Changes:
 * Remove the agent of iRobot Create, and replaced with KUKA Youbot.

Motivation:
 * Since iRobot Create is basically the same function as pioneer, just without camera. It's not necessary to keep it.

 * Consider that simulation of pr2 and icub_with_hands are quite slow, we may want an agent that could do both locomotion and grasping/placing job, and have a simpler structure so that the simulation could be faster. 

 So I added KUKA youbot (https://github.com/youbot/). Its simulation is 5~6 times faster than PR2 (60-70 steps/s on my laptop).

youbot:
image

youbot doing goal task trained by SAC:
youbot


Jialn:
> If create is replaced, please make sure to update all the places using or mentioning irobot create throughout the whole project. For example:
> 
> https://github.com/HorizonRobotics/SocialRobot/blob/1bfac7bab9779e3126034c8e63142845a3daa534/python/social_bot/envs/grocery_ground.py#L522-L523
> 
> in groucery_ground.py need to be updated.

Comments and Read-me and tests are also updated.

Jialn:
> Nice addition Jiangtao. I didn't know that irobot doesn't have touch sensor. May worth leaving it in if we have plans to add touch sensor in the future.

irobot does not have a touch sensor, but have a gazebo ray sensor (infrared range sensor) which serves similar function. We can also add that easily to pioneer.

Actually model of irobot create is inside gazebo's model database and we did not keep a copy in our repo. So you can see I barely remove anything. It's very easy to add it back by just adding a simple config in agent_cfg.jason if needed in the future.

PR #85, opened 2019-10-21
some code cleaning and refactoring

Jialn:
I have put all tasks to grocery_ground.py, and made it too complex. Also some redundant codes structures are left. Before we implement new features like embodied teacher, let's do some code cleaning first. 

Changes:
- remove GroceryGroundTaskBase class, it's redundant. The initializing operations of it are merged to the base task class.
- move the tasks from grocery_ground.py to teacher_tasks.py
- the fixed configs for grocery env by registering gym env name are barely used, and can be replaced with more flexible gin configurations, so removed;
- auto detect camera sensor's existing in unit-test; 
- add several training curves so that we can have comparison latter;

Some codes are changed due to pre-commit auto formatting. If you have suggestions feel free to tell me.


python/social_bot/teacher_tasks.py
```diff
@@ -205,3 +208,462 @@ def set_goal_name(self, goal_name):
         logging.debug('Setting Goal to %s', goal_name)
         self._goal_name = goal_name
 
+
+@gin.configurable
+class GroceryGroundGoalTask(GoalTask):
```
le-horizon: Not necessary to address this in this PR, but this task doesn't contain any reference to the GroceryGround env.  Does it make sense to call this something like MultiGoalWithDistractionTask?

Haichao-Zhang: Not sure if it is a good choice to directly move all the tasks from grocery_ground.py to teacher_tasks.py. As after doing this, some of the tasks in teacher_tasks.py are highly coupled with some specific world/env.
The ideal case is to only keep relatively general tasks that can be reused in different worlds or environment in teacher_tasks.py while leaving those coupled with specific env into separate files.

Jialn: Yes, coupled with specific env tasks should leave to the env file. I was considering to decouple the 3 tasks with grocery env. Goal task and kicking-ball task could be more general. ICubAuxTask is also supposed to be a general helper task providing reward-shaping for walking by 2 legs.
I was also considering rename GroceryGround to PlayGround and remove it's dependency on specific task. So that we can assemble different tasks and different morphology of teachers in the playground through gin. Too many changes in one pull request are hard to review. Let's open another PR for these.

Jialn: Good point. I was considering decouple the tasks with specific env. Goal task and kicking-ball task are supposed to more general. I was also considering rename GroceryGround to PlayGround and remove it's dependency on specific task. So that we can assemble different tasks and different morphology of teachers in the playground through gin. 
Too many changes in one pull request are hard to review. Let's open another PR for these.

Haichao-Zhang: Sounds good. We can do this progressively.

PR #86, opened 2019-10-22
decouple tasks from grocery env and classes renaming

Jialn:
As proposed in #85 , changes are:
1. Rename GroceryGround to PlayGround and remove it's dependency on specific task. So that we can assemble different tasks and different morphology of teachers(to do) in the playground through gin. 
2. Decouple the 3 tasks with grocery ground env. Goal task and kicking-ball task could be more general. ICubAuxTask is also supposed to be a general helper task providing reward-shaping for walking by 2 legs. But currently they still have some requirements on the world file: the world must have a model "ball' for the kicking-ball task and goal task, and the name of ground link in the world file should be "ground_plane" because icub_aux_task's contact-sensor require the link name.
3. Unify the names of gin files, and remove some distractive models in the world file by default for the GoalWithDistractionTask.


python/social_bot/envs/play_ground.py
```diff
@@ -294,8 +281,6 @@ def _get_camera_observation(self):
             image = PIL.Image.fromarray(image).resize(self._resized_image_size,
                                                       PIL.Image.ANTIALIAS)
             image = np.array(image, copy=False)
-        if self._data_format == "channels_first":
```
Jialn: channel problem of tf2.0 is fixed, so remove this option

python/social_bot/envs/play_ground.py
```diff
@@ -67,103 +66,101 @@ class GroceryGround(GazeboEnvBase):
...
-        if agent_type.find('icub') != -1:
-            icub_aux_task_group = TaskGroup()
-            icub_standing_task = ICubAuxiliaryTask(step_time=step_time)
-            icub_aux_task_group.add_task(icub_standing_task)
-            self._teacher.add_task_group(icub_aux_task_group)
+        main_task = task(step_time=step_time)
+        task_group = TaskGroup()
+        task_group.add_task(main_task)
+        self._teacher.add_task_group(task_group)
+        if secondary_task != None:
```
Jialn: Not sure if this is a good way to support multiple tasks

python/social_bot/teacher_tasks.py
```diff
@@ -226,6 +226,7 @@ def __init__(self,
                  random_range=10.0,
                  random_goal=False,
                  sparse_reward=True,
+                 step_time=0.1,
```
Jialn: actually step_time is not used in this GoalWithDistractionTask task. but other tasks may need playground to tell them step_time. Any better way for this?

python/social_bot/teacher_tasks.py
```diff
@@ -274,10 +276,6 @@ def __init__(self,
             percent_full_range_in_curriculum=percent_full_range_in_curriculum,
             max_reward_q_length=max_reward_q_length)
         self._random_goal = random_goal
-        self._objects_in_world = [
-            'placing_table', 'plastic_cup_on_table', 'coke_can_on_table',
-            'hammer_on_table', 'cafe_table', 'ball'
-        ]
```
le-horizon: Is this deletion intentional?

Jialn: yes, this makes the task depends on world file, besides
 the default world file has changed to playground.world, so removed

python/social_bot/envs/play_ground.py
Jialn: good idea

python/social_bot/teacher_tasks.py
le-horizon: I see.  This changes the setup of the task though, so at least indicate in the commit message that this is happening.

python/social_bot/teacher_tasks.py
Jialn: Nevermind, we can keep it this way. interface of tasks should be more unified.
Because sim step time and substeps might be different for each agent and env, and can be override in gin config, in some task this option is used to get the time span of 1 step from the env (e.g., reward shaping, caculate normalized speed to goal from distance to goal of 2 steps will need this).  So we need something like Task(stepsize=**) in env. But not all env need this.

python/social_bot/teacher_tasks.py
Jialn: OK, I'll add the message when checking in.

python/social_bot/teacher_tasks.py
le-horizon: Thanks for explaining Jiangtao! It seems strange to have a configurable named step_time used in computing speed.  Shouldn't speed be fully determined by the physics?  Maybe consider changing the name of this variable, and maybe also the formula to make it more natural?

Jialn: step_time is the time passed by an env step. 
`step_time = substeps * substep_time`
substep_time is the simulator's accuracy, i.e., max_step_size in world file. Typically step_time is 0.1 (10Hz roll-out freq), and max_step_size is 0.001. 

While speed is numerical normalized, since there are 100 simulator sub-steps depends on the configuration, speed to goal (the agent speed projected on the direction to goal) calculated by simulator is from the last 0.001s sub-step, does not reflect the real progress made to goal in the entire 0.1s env step, especially when the agent is in an accelerating progress. (distance_to_goal-pre_distance_to_goal)/step_time is more accurate here.

```
normalized_progress = (distance_to_goal - pre_distance_to_goal) / initial_distance
speed_to_goal =  (distance_to_goal - pre_distance_to_goal) / step_time
```
I'll add some comments for this.

le-horizon: I see, step_time and substeps are configured, and substep_time is inferred from the two?

If step_time is the elapsed time per each env step, then shouldn't it be part of the env, instead of the task?

Jialn: yes.  substeps is inferred from the step_time and substep_time. it is part of the env, but the task need this. So when create the task, this parameter should be passed.

python/social_bot/envs/play_ground.py
```diff
@@ -67,103 +68,101 @@ class GroceryGround(GazeboEnvBase):
...
+                 world_name="play_ground.world",
+                 task=GoalWithDistractionTask,
+                 secondary_task=None,
                  with_language=False,
                  use_image_observation=False,
                  image_with_internal_states=False,
-                 task_name='goal',
-                 agent_type='pioneer2dx_noplugin',
                  world_time_precision=None,
                  step_time=0.1,
```
Jialn: world_time_precision is the substep time. it is read from agent.cfg by default. 

step_time=0.1 by default suites for most agent/task (10Hz control rate). But tasks need higher control frequency prefer lower step_time. e.g., iCub is trained with 0.05,  a steptime of 0.1 on the icub walking task won't be success).

python/social_bot/envs/play_ground.py
```diff
@@ -67,103 +68,101 @@ class GroceryGround(GazeboEnvBase):
...
-        wf_path = os.path.join(social_bot.get_world_dir(),
-                               "grocery_ground.world")
-        with open(wf_path, 'r+') as world_file:
+        wd_path = os.path.join(social_bot.get_world_dir(), world_name)
+        with open(wd_path, 'r+') as world_file:
             world_string = self._insert_agent_to_world_file(
                 world_file, agent_type)
         if world_time_precision is None:
             world_time_precision = agent_cfg['max_sim_step_time']
         sub_steps = int(round(step_time / world_time_precision))
```
Jialn: number of sub_steps is calculated here.

python/social_bot/envs/play_ground.py
```diff
@@ -67,103 +68,101 @@ class GroceryGround(GazeboEnvBase):
...
-
-        main_task_group = TaskGroup()
-        main_task_group.add_task(main_task)
-        self._teacher.add_task_group(main_task_group)
-        if agent_type.find('icub') != -1:
-            icub_aux_task_group = TaskGroup()
-            icub_standing_task = ICubAuxiliaryTask(step_time=step_time)
-            icub_aux_task_group.add_task(icub_standing_task)
-            self._teacher.add_task_group(icub_aux_task_group)
+        main_task = task(step_time=step_time)
```
Jialn: pass step_time to task here.

python/social_bot/teacher_tasks.py
Haichao-Zhang: Since max_step_size is specified in the world file, and NUM_SIMULATION_STEPS (substeps) is typically specified in specific tasks, it is indeed better to compute the step_time value based on these two variables. This calculation can be done outside of this task as you mentioned.
For now, perhaps set the default value to an invalid value and add assertion on value when using it?

Jialn: There is a maximal max_step_size for each agent. for example, pr2 requires max_step_size <= 0.001 for gripping task. iCub also require max_step_size <= 0.001 for walking task. otherwise you can't train a successful policy.
But pioneer works well when max_step_size=0.005. The performance for pioneer could be 5 times faster, if we set it to 0.005 rather than 0.001.

So max_step_size will differ according agents for performance reason. Currently we read max_step_size, i.e., world_time_precision, from the file agent.cfg. 

It is more natural to define step_time rather than NUM_SIMULATION_STEPS (substeps) for a task. Otherwise when agent changes, max_step_size changes, and then the time span of env step will be changed.
for example, if we set NUM_SIMULATION_STEPS (substeps)  to 100;  PR2-> pioneer; then max_step_size 0.001->0.005; control frequency 10Hz -> 2Hz 
If given steptime 0.1, then the control frequency will be fixed to 10Hz. NUM_SIMULATION_STEPS (substeps) will be caculated by steptime and max_step_size of agent.

So the model determines the max_step_size(i.e., substep_time/world_time_precision); and we should set steptime according to the task; it is indeed better to compute the NUM_SIMULATION_STEPS (substeps) from the two variables.

le-horizon: Thanks for the clarifications Jiangtao.  Since step_time is the elapsed time per each step, and env class (playground) already has step_time, we should probably remove step_time from being a class member variable of the Tasks.  We can send in a world obj to task_specific_observation() function to compute speed.  The rest of the class probably doesn't need to know about step_time?

Jialn: Thanks for the kindly suggestion! step_time is removed from being a class member variable of the Tasks now. I added a get_step_time() function to environment. Please review again :-)

Jialn:
Made some changes to unify  the names of gin files. 
Also change the name of object_to_insert to distraction_list in GoalWithDistractionTask and make it gin configurable. For the default distraction_list I removed hammer and bookshelf as the default name. hammer is too short basically has no effect on agent, and bookshelf block too much view which makes the agent hard to see the goal when nearby it.

python/social_bot/teacher_tasks.py
```diff
@@ -580,31 +579,23 @@ def __init__(self,
...
         self._goal_name = 'goal'
-        self._success_distance_thresh = success_distance_thresh,
-        self._objects_in_world = [
-            'placing_table', 'plastic_cup_on_table', 'coke_can_on_table',
-            'hammer_on_table', 'cafe_table', 'ball'
-        ]
-        self._step_time = step_time
+        self._success_distance_thresh = success_distance_thresh
         self._target_speed = target_speed
         self.reward_weight = reward_weight
```
le-horizon: pass reward_weight to super(), instead of having a separate variable in this derived class.

python/social_bot/teacher_tasks.py
```diff
@@ -580,31 +579,23 @@ def __init__(self,
...
-            step_time (float): used to caculate speed of the agent
             reward_weight (float): the weight of the reward
         """
-        GoalTask.__init__(
-            self,
+        super().__init__(
             max_steps=max_steps,
             goal_name=goal_name,
             fail_distance_thresh=fail_distance_thresh,
             random_range=random_range)
```
le-horizon: Pass reward_weight to super(), since it's already in GoalTask (or Task if you choose to put it there).

python/social_bot/teacher_tasks.py
```diff
@@ -580,31 +579,23 @@ def __init__(self,
...
         self._goal_name = 'goal'
-        self._success_distance_thresh = success_distance_thresh,
-        self._objects_in_world = [
-            'placing_table', 'plastic_cup_on_table', 'coke_can_on_table',
-            'hammer_on_table', 'cafe_table', 'ball'
-        ]
-        self._step_time = step_time
+        self._success_distance_thresh = success_distance_thresh
```
le-horizon: pass success_distance_thresh to super() instead of assigning here.

Jialn:
> Hey Jiangtao,
> 
> Sorry for taking long to review. I took a closer look, and included some comments to hopefully improve the quality of the code.
> 
> Thanks,
> Le

Thanks for the careful reviews. Very helpful for the code quality.

python/social_bot/teacher_tasks.py
```diff
@@ -244,6 +249,7 @@ class GroceryGroundGoalTask(GoalTask):
     def __init__(self,
                  max_steps=500,
                  goal_name="ball",
+                 distraction_list=['coke_can', 'table', 'car_wheel', 'plastic_cup', 'beer'],
```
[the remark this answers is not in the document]
Jialn: Yes, that's due to a historical reason that in the setting before, default goal name might not in the self._distraction_list.  It's better to add 'ball' into the list now.

python/social_bot/teacher_tasks.py
```diff
@@ -297,43 +304,37 @@ def __init__(self,
...
-        if GoalTask.should_use_curriculum_training(self):
+        if self.should_use_curriculum_training():
             logging.info("start_range %f, reward_thresh_to_increase_range %f",
                          self._start_range,
                          self._reward_thresh_to_increase_range)
         self._pos_list = list(itertools.product(range(-5, 5), range(-5, 5)))
         self._pos_list.remove((0, 0))
         self.reward_weight = reward_weight
-        self.task_vocab += self._objects_in_world + self._objects_to_insert
+        self.task_vocab += [goal_name] + self._distraction_list
```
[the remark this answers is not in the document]
Jialn: A fix to historical problem. The setting before was that default goal is from world file and might not in the self._distraction_list. 
We can add default ball to the list and so do not need to add goal_name to vocab.

python/social_bot/teacher_tasks.py
```diff
@@ -91,6 +94,7 @@ def __init__(self,
         self._use_curriculum_training = use_curriculum_training
         self._start_range = start_range
         self._is_full_range_in_curriculum = False
+        self.reward_weight = reward_weight
```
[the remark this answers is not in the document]
Jialn: Fixed now. 
Also a historical problem. reward_weight should be a part of teacher.Task. 
reward_weight of each task is multiplied in "teach()" in "teach.py"

python/social_bot/teacher.py
```diff
@@ -61,13 +61,14 @@ def __init__(self):
         self._agent_name = None
         self.reward_weight = 1.0
 
-    def setup(self, world, agent_name):
+    def setup(self, world, agent_name, env):
```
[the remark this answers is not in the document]
Jialn: yes, env is enough

python/social_bot/envs/play_ground.py
```diff
@@ -67,104 +68,96 @@ class GroceryGround(GazeboEnvBase):
...
-                note that 'agent_type' should be the same str as the model's name
-            world_time_precision (float|None): if not none, the time precision of
-                simulator, i.e., the max_step_size defined in the agent cfg file, will be
-                override. e.g., '0.002' for a 2ms sim step
-            step_time (float): the peroid of one step of the environment.
+            world_time_precision (float|None): this parameter depends on the agent. 
+                if not none, the default time precision of simulator, i.e., the max_step_size
+                defined in the agent cfg file, will be override. Note that pr2 and iCub
+                requires a max_step_size <= 0.001, otherwise cannot train a successful policy.
+            step_time (float): the peroid of one step() function of the environment in simulation.
```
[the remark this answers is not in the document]
Jialn: Done

python/social_bot/teacher_tasks.py
```diff
@@ -407,31 +406,31 @@ def setup(self, env):
...
                 os.path.join(social_bot.get_model_dir(), "agent_cfg.json"),
                 'r') as cfg_file:
             agent_cfgs = json.load(cfg_file)
         self._joints = agent_cfgs[self._agent_type]['control_joints']
 
-    def run(self, agent, world):
+    def run(self):
         """
         Start a teaching episode for this task.
         Args:
```
le-horizon: remove comments

python/social_bot/teacher_tasks.py
```diff
@@ -614,19 +613,19 @@ def setup(self, env):
         time.sleep(0.2)
         self._world.step(20)
 
-    def run(self, agent, world):
+    def run(self):
         """
         Start a teaching episode for this task.
         Args:
             agent (pygazebo.Agent): the learning agent
```
le-horizon: remove comments

python/social_bot/teacher_tasks.py
```diff
@@ -128,23 +128,23 @@ def _push_reward_queue(self, value):
     def get_random_range(self):
         return self._random_range
 
-    def run(self, agent, world, distractions=None):
+    def run(self, distractions=None):
         """
         Start a teaching episode for this task.
         Args:
             agent (pygazebo.Agent): the learning agent
```
le-horizon: remove comments and add comment on distractions

Jialn:
Hi Le, 
Comments has been fixed. Seems need your approve again... Thanks.

le-horizon:
Thanks for the fixes Jiangtao!  Approved and Merged.

python/social_bot/envs/play_ground.py
le-horizon: Does this mean PlayGround will always output in the format "channels_last"?

Jialn:
> Btw, forgot that you'll also need to change the relevant gin files in alf for the PlayGround env name update, and image_format API removal.
> 
> Thanks,
> Le

Currently examples of ALF do not have gin file for GroceryGround/PlayGround env.

Jialn:
> Does this mean PlayGround will always output in the format "channels_last"?

Yes... This was bought in because there's a bug for the earlier version tf2.0-alpha-gpu that only channels_first can be supported while tf2.0-alpha-cpu can only support channels_last. So we have to switch it when using different local PC or servers. It's fixed now. I thought this was of no use. Do we still need this option?

le-horizon:
I see.  Sounds good then.  If we need it later, we can always add it back.  Just need to remember that we removed channel order from playground.

PR #87, opened 2019-10-22
Pr discourage hitting distractions

le-horizon:
Hey Jiangtao and Wei,

I'm trying to discourage the bot from bumping into random objects when chasing after the goal target.

This change reduces reward by a configurable value (distraction_penalty) every time the robot bumps into an object (gets very close to an object).  It also decreases the reward_q average, and prevents the curriculum from advancing if robot keeps bumping into distractions.

Let me know what do you think about this change.  I started a training session with penalty distance threshold = 0.25 (compared to success distance of 0.5) and penalty = 0.5 (compared to reward of reaching goal 1.0).

Thanks,
Le

### DIFF
--- a/python/social_bot/teacher_tasks.py
+++ b/python/social_bot/teacher_tasks.py
@@ -65,7 +65,8 @@ class GoalTask(teacher.Task):
                 it's considered a failure and is given reward -1
             distraction_penalty_distance_thresh (float): if positive, penalize agent getting too close
                 to distraction objects (objects that are not the goal itself)
-            distraction_penalty (float): how much to penalize getting too close to distraction objects
+            distraction_penalty (float): positive float of how much to penalize getting too close to
+                distraction objects
             sparse_reward (bool): if true, the reward is -1/0/1, otherwise the 0 case will be replaced
                 with normalized distance the agent get closer to goal.
             random_range (float): the goal's random position range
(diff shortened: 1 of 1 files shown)


python/social_bot/teacher_tasks.py
```diff
@@ -308,7 +332,8 @@ def run(self, agent, world):
         if self._random_goal:
             random_id = random.randrange(len(self._goals))
             self.set_goal_name(self._goals[random_id])
-        yield from GoalTask.run(self, agent, world)
+        yield from GoalTask.run(self, agent, world,
+            distractions=self._objects_in_world + self._objects_to_insert)
```
Jialn: objects_in_world has been removed...

python/social_bot/teacher_tasks.py
```diff
@@ -46,6 +46,8 @@ def __init__(self,
                  goal_name="goal",
                  success_distance_thresh=0.5,
                  fail_distance_thresh=0.5,
+                 distraction_penalty_distance_thresh=0,
+                 distraction_penalty=0,
```
Jialn: Since distraction_penalty_distance_thresh act as the ON/OFF for this, maybe we can set a default recommended distraction_penalty?

le-horizon: Good point, done.

le-horizon:
Oops, looks like the commit to address comments revoked Jiangtao's approval..
Jiangtao or Wei, could check and approve again?

Thanks,
Le

PR #91, opened 2019-10-29
Task code refactoring 2

Jialn:
Changes:

- merge Task.setup() to Task._init(); it is safe to insert models during init now.
- move Task from teacher.py to teacher_tasks.py
- rename teacher_tasks.py to tasks.py


python/social_bot/envs/gazebo_base.py
```diff
@@ -133,6 +133,48 @@ def render(self, mode='human'):
...
+                <uri>model://""" + model + """</uri>
+            </include>
+            <pose frame=''>""" + pose + """</pose>
+        </model>
+        </sdf>
+        """
+        self._world.insertModelFromSdfString(model_sdf)
+        # Sleep for a while waiting for gzserver to finish the inserting
+        # operation. Or it may not be successfully inserted and report error.
+        time.sleep(0.2)
```
le-horizon: Not directly related to this change: If goal is not reached in max_steps (100 or 200 in some tasks), then we'll have to wait 0.2 second?  Would that be too much overhead?  Would it be possible to step first, and catch the exception if it occurs?  Also, have anyone tested with fewer than 20 steps to save some time?

python/social_bot/envs/gazebo_base.py
```diff
@@ -133,6 +133,48 @@ def render(self, mode='human'):
...
+            model_list (list) : the list of the models
+        """
+        obj_num = len(model_list)
+        for obj_id in range(obj_num):
+            model_name = model_list[obj_id]
+            if self._world.model_list_info().find(model_name) == -1:
+                self._world.insertModelFile('model://' + model_name)
+                logging.debug('model ' + model_name + ' inserted')
+                time.sleep(0.2)
+                self._world.step(20)
```
le-horizon: This is sleeping and stepping for every model inserted into the world?  Probably only need to do that after all models are inserted?  Only potential problem would be if the model list contains duplicates, then the same item will be inserted twice?  But that's probably a problem of the input list?

python/social_bot/tasks.py
```diff
@@ -43,6 +103,7 @@ class GoalTask(teacher.Task):
     """
 
     def __init__(self,
+                 env,
                  max_steps=500,
```
le-horizon: In the future, maybe we could put max_steps into Task, and put the main while loop in run into Task.run, and have a more modularized structure for different tasks?  Maybe define a task_success() function, etc..  For this change, no need to do that.

python/social_bot/envs/gazebo_base.py
Jialn: It's fine. Only during initializing will there be inserting operation. Reset
 does not need that.
I have spend a lot time on the insertion problem, increase the sleeping and stepping little by little... this is safe for most case.

le-horizon: Sounds good, thanks for explaining!

python/social_bot/envs/gazebo_base.py
Jialn: yes, it's a must. Each insertion need to wait for some time or gazebo crashes....
There is a checking for duplicates, line 173 
if self._world.model_list_info().find(model_name) == -1

python/social_bot/tasks.py
Jialn: thanks for the suggestin. Should have a unified modular structure for tasks.

python/social_bot/envs/gazebo_base.py
Jialn: but this still is buggy. Like, if we instead an appletree first, we can not insert a tree latter..

PR #93, opened 2019-10-30
merge GoalWithDistractionTask with GoalTask

Jialn:
Since distraction_lists and its penalty are actually processed in GoalTask, now GoalWithDistractionTask only do things like select random goal if random_goal is enabled, it would be better to remove GoalWithDistractionTask and merge the function to GoalTask.

### DIFF
--- a/python/social_bot/envs/simple_navigation.py
+++ b/python/social_bot/envs/simple_navigation.py
@@ -91,7 +91,9 @@ class SimpleNavigation(GazeboEnvBase):
         task_group = teacher.TaskGroup()
         task = GoalTask(
             env=self,
+            max_steps=120,
             goal_name="goal",
+            distraction_list=[],
             random_range=2.0)
         task_group.add_task(task)
         self._teacher.add_task_group(task_group)
--- a/python/social_bot/tasks.py
+++ b/python/social_bot/tasks.py
@@ -205,11 +205,9 @@ class GoalTask(Task):
     def get_random_range(self):
         return self._random_range
 
-    def run(self, distractions=None):
+    def run(self):
         """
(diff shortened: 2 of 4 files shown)


python/social_bot/tasks.py
```diff
@@ -104,14 +104,16 @@ class GoalTask(Task):
 
     def __init__(self,
                  env,
-                 max_steps=500,
+                 max_steps=200,
```
Jialn: Default max_steps of Goaltask is changed to 200 which is more reasonable for the size of playground. 500 seems too large.

python/social_bot/envs/simple_navigation.py
```diff
@@ -91,7 +91,9 @@ def __init__(self,
         task_group = teacher.TaskGroup()
         task = GoalTask(
             env=self,
+            max_steps=120,
```
Jialn: Default max_steps for simple-navigation whose range is smaller has been set to 120.

python/social_bot/tasks.py
```diff
@@ -156,21 +160,29 @@ def __init__(self,
         self._use_curriculum_training = use_curriculum_training
         self._start_range = start_range
         self._is_full_range_in_curriculum = False
+        self._random_goal = random_goal
+        self._distraction_list = distraction_list
+        self._object_list = distraction_list
+        if not goal_name in distraction_list:
+            self._object_list.append(goal_name)
+        self._goals = self._object_list
```
Jialn: Let's use distraction_list + [goal_name] as the range of goals if random_goal is enabled for now. 
If specifying the list for goals is needed, we can define goal_name as a list.

PR #97, opened 2019-11-01
add all used gazebo model files into repository so we don't need to f…

le-horizon:
…etch them from gazebosim.org, and we can maintain stability of these models.

These are all the models I can find from the current repo.
We probably need to check it every time we add a new model to SocialRobot.

Thanks,
Le


le-horizon:
> Please add some description about the source of these files
> Add make sure to we follow the license requirement: https://bitbucket.org/osrf/gazebo_models/src/default/LICENSE
> 
> Is it possible to disable downloading from the internet?

Good points as always @emailweixu.  Done both.  PTAL?

Thanks,
Le

python/social_bot/models/database.config
```diff
@@ -1,13 +1,26 @@
 <?xml version='1.0'?>
 <database>
-  <name>Models used by SocialBot</name>
+  <name>Models used by SocialBot, copied or adapted from https://bitbucket.org/osrf/gazebo_models</name>
```
emailweixu: not all of them are from https://bitbucket.org/osrf/gazebo_models.
 ball, texture_ground_plane, cartpole are created by us
 pr2 and youbot_noplugin are from somewhere else

Jialn: pr2 and yobot are also copied from there...  iCub is from https://github.com/robotology/icub-gazebo

Jialn:
Thanks for the work. It's more convient now. There's problem download these models in mainland China I have to copy model file by hand everytime setup the running environment.

le-horizon:
Glad to know it's helpful, Jiangtao.  I updated the comments, let me know if you want to make any changes to where the model files are from.

Thanks,
Le

PR #101, opened 2019-11-08
Pr move distractions closer in curriculum

le-horizon:
This includes several changes to allow adjusting goal's direction in curriculum, allow resetting and picking new random goal in one episode, moving distraction objects also according to curriculum, using polar coordinate in randomly setting goal position.

### DIFF
--- a/python/social_bot/envs/play_ground.py
+++ b/python/social_bot/envs/play_ground.py
@@ -208,7 +208,7 @@ class PlayGround(GazeboEnvBase):
             If with_language, it is a dictionary with key 'data' and 'sentence'
             If not with_language, it is a numpy.array or image for observation
         """
-        if self._with_language:
+        if False and self._with_language:
             sentence = action.get('sentence', None)
             if type(sentence) != str:
                 sentence = self._teacher.sequence_to_sentence(sentence)
--- a/python/social_bot/envs/play_ground.py
+++ b/python/social_bot/envs/play_ground.py
@@ -219,9 +219,7 @@ class PlayGround(GazeboEnvBase):
             controls = action['control']
         else:
             sentence = ''
-            action_ctrl = action
-        controls = np.clip(action_ctrl, -1.0, 1.0) * self._agent_control_range
-        controls = dict(zip(self._agent_joints, controls))
+            controls = action
         self._agent.take_action(controls)
         self._world.step(self._sub_steps)
         teacher_action = self._teacher.teach(sentence)
--- a/python/social_bot/envs/play_ground.py
+++ b/python/social_bot/envs/play_ground.py
@@ -215,7 +215,13 @@ class PlayGround(GazeboEnvBase):
             controls = action['control']
         else:
             sentence = ''
-            controls = action
+            action_ctrl = action
+        task = (self._teacher._current_task_group
+            ).get_current_task_non_generator()
+        if isinstance(task, GoalTask) and task._also_curriculum_control:
+            action_ctrl *= task._control_range
+        controls = np.clip(action_ctrl, -1.0, 1.0) * self._agent_control_range
+        controls = dict(zip(self._agent_joints, controls))
         self._agent.take_action(controls)
         self._world.step(self._sub_steps)
         teacher_action = self._teacher.teach(sentence)
--- a/python/social_bot/envs/play_ground.py
+++ b/python/social_bot/envs/play_ground.py
@@ -72,6 +72,7 @@ class PlayGround(GazeboEnvBase):
                  world_name="play_ground.world",
                  tasks=[GoalTask],
                  with_language=False,
+                 with_agent_language=False,
                  use_image_observation=False,
                  image_with_internal_states=False,
                  max_steps=200,
@@ -92,6 +93,7 @@ class PlayGround(GazeboEnvBase):
                 grocery_ground.world
             tasks (list): a list of teacher.Task, e.g., GoalTask, KickingBallTask
             with_language (bool): The observation will be a dict with an extra sentence
+            with_agent_language (bool): Include agent sentence in action space.
             use_image_observation (bool): Use image, or use low-dimentional states as
                 observation. Poses in the states observation are in world coordinate
             max_steps (int): episode will end in so many steps, will be passed to the tasks
@@ -123,6 +125,7 @@ class PlayGround(GazeboEnvBase):
         self._action_cost = action_cost
         self._with_language = with_language
         self._seq_length = vocab_sequence_length
+        self._with_agent_language = with_agent_language
         self._use_image_obs = use_image_observation
         self._image_with_internal_states = self._use_image_obs and image_with_internal_states
 
@@ -154,6 +157,7 @@ class PlayGround(GazeboEnvBase):
             agent_type=agent_type,
             config=agent_cfg,
             with_language=with_language,
+            with_agent_language=with_agent_language,
             vocab_sequence_length=self._seq_length,
             use_image_observation=use_image_observation,
             resized_image_size=resized_image_size,
@@ -208,7 +212,7 @@ class PlayGround(GazeboEnvBase):
             If with_language, it is a dictionary with key 'data' and 'sentence'
             If not with_language, it is a numpy.array or image for observation
         """
-        if False and self._with_language:
+        if self._with_agent_language and self._with_language:
             sentence = action.get('sentence', None)
             if type(sentence) != str:
                 sentence = self._teacher.sequence_to_sentence(sentence)
--- a/python/social_bot/envs/play_ground.py
+++ b/python/social_bot/envs/play_ground.py
@@ -216,10 +216,6 @@ class PlayGround(GazeboEnvBase):
         else:
             sentence = ''
             action_ctrl = action
-        task = (self._teacher._current_task_group
-            ).get_current_task_non_generator()
-        if isinstance(task, GoalTask) and task._also_curriculum_control:
-            action_ctrl *= task._control_range
         controls = np.clip(action_ctrl, -1.0, 1.0) * self._agent_control_range
         controls = dict(zip(self._agent_joints, controls))
         self._agent.take_action(controls)
--- a/python/social_bot/tasks.py
+++ b/python/social_bot/tasks.py
@@ -148,6 +148,7 @@ class GoalTask(Task):
                  also_curriculum_distractions=True,
                  also_curriculum_target_angle=False,
                  also_curriculum_control=False,
+                 switch_goal_within_episode=False,
                  start_range=0,
                  increase_range_by_percent=50.,
                  reward_thresh_to_increase_range=0.4,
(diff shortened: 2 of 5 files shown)


python/social_bot/envs/play_ground.py
```diff
@@ -177,7 +177,7 @@ def __init__(self,
             high=1.0,
             shape=[len(self._agent_joints)],
             dtype=np.float32)
-        if self._with_language:
+        if False and self._with_language:
```
Jialn: Why is here a "False" ?

python/social_bot/envs/play_ground.py
```diff
@@ -235,14 +235,18 @@ def step(self, action):
...
             sentence = action.get('sentence', None)
             if type(sentence) != str:
                 sentence = self._teacher.sequence_to_sentence(sentence)
             action_ctrl = action['control']
         else:
             sentence = ''
             action_ctrl = action
+        task = (self._teacher._current_task_group
+            ).get_current_task_non_generator()
+        if isinstance(task, GoalTask) and task._also_curriculum_control:
```
Jialn: task specific control_range might not be widely used for many tasks. I think we can override env.agent_control_range in GoalTask, or add a set_control_range to playground and use env.set_control_range() in GoalTask

python/social_bot/teacher.py
```diff
@@ -122,13 +123,17 @@ def reset(self, agent, world):
...
             tid = random.randint(0, len(self._tasks) - 1)
+            self._current_tid = tid
             self._current_task = self._tasks[tid].run()
             self._current_reward_weight = self._tasks[tid].reward_weight
             # This send will cause self._current_task to execute until the first
             # yield. We ignore the first yielded value.
             self._current_task.send(None)
         return self._current_task
 
+    def get_current_task_non_generator(self):
```
Jialn: what does "_non_generator" mean here?

python/social_bot/tasks.py
```diff
@@ -115,6 +116,9 @@ def __init__(self,
                  random_range=5.0,
                  random_goal=False,
                  use_curriculum_training=False,
+                 also_curriculum_distractions=True,
+                 also_curriculum_target_angle=False,
+                 also_curriculum_control=False,
```
Jialn: also_curriculum_control_range, may be remove "also"?

python/social_bot/tasks.py
```diff
@@ -139,6 +143,10 @@ def __init__(self,
             random_range (float): the goal's random position range
             random_goal (bool): if ture, teacher will randomly select goal from the object list each episode
             use_curriculum_training (bool): when true, use curriculum in goal task training
+            also_curriculum_distractions (bool): move distractions according to curriculum as well
+            also_curriculum_target_angle (bool): enlarge angle to target when initializing target according
+                to curriculum.  Only when all angles are satisfied does curriculum try to increase distance.
+                Uses range of 0-180 degrees, starting from 60 with increments of 15.
```
Jialn: comment for also_curriculum_control_range

python/social_bot/tasks.py
```diff
@@ -196,10 +220,26 @@ def _push_reward_queue(self, value):
...
+                logging.info("Raising random_angle to %d", self._random_angle)
+            if (not self._also_curriculum_target_angle or
+                self._random_angle > 360):
+                self._random_angle = 60
+                new_range = min(
+                    (1. + self._increase_range_by_percent
+                        ) * self._random_range,
+                    self._orig_random_range)
+                new_control_range = min(1.0,
+                    (1. + self._increase_range_by_percent) *
```
Jialn: Any experiments on increasing control_range? From my experiences, at the start of training progress actions are more random, so small ranges make the exploration not efficient. (Pure random actions have a (v*t)^(1/2) expectation explore range, v is the average speed).  A higher control range should be at start.  So I've already set a relatively larger initial action range in the "effort_limit" term of joints in model file. There is also an optional action_cost argument in playground to encourage the agent using limited actions as possible while completing its main task. As it is already higher enough in the beginning, In my view increasing control_range might not be so meaningful.

Also, as the magnitude of actions changes in (s, a, s', r), it might have unknown affect on off-policy RL algorithm. MDP become POMDP, I'm not quite sure but theoretically we may need to add new_control_range to observation.

python/social_bot/tasks.py
```diff
@@ -269,16 +301,63 @@ def run(self):
...
+                obj = self._world.get_agent(obj_name)
+                if not obj:
+                    continue
+                obj_loc, _ = obj.get_pose()
+                obj_loc = np.array(obj_loc)
+                distraction_dist = np.linalg.norm(agent_loc - obj_loc)
+                if (distraction_dist >=
+                    self._distraction_penalty_distance_thresh):
+                    continue
+                if obj_name == self._goal_name and dot > 0.707:
```
Jialn: "dot > 0.707" means there will be a distracting penalty if closer to the goal if the angel is less than 45°, is this on purpose? Some camera horizontal view is larger than 90 degree, and in pure state observation case agent direction to the ball is not so important. Maybe we can ignore the dot here.

python/social_bot/tasks.py
```diff
@@ -269,16 +301,63 @@ def run(self):
...
+        distractions = OrderedDict()
+        for item in self._distraction_list:
+            if item is not self._goal_name:
+                distractions[item] = 1
+        if len(distractions) and self._also_curriculum_distractions:
+            rand_id = random.randrange(len(distractions))
+            distraction = self._world.get_agent(list(distractions.keys())[rand_id])
+            self._move_goal_impl(distraction, agent_loc, agent_dir)
+
+    def _move_goal_impl(self, goal, agent_loc, agent_dir):
```
Jialn: what does "_impl" mean here?

python/social_bot/tasks.py
```diff
@@ -290,6 +369,12 @@ def _move_goal(self, goal, agent_loc):
         while True:
             loc = (random.random() * range - range / 2,
                    random.random() * range - range / 2, 0)
+            if self._also_curriculum_target_angle and not self._is_full_range_in_curriculum:
+                dist = random.random() * range
+                angle = math.radians(
+                    math.degrees(agent_dir[2]) +
+                    random.random() * self._random_angle - self._random_angle / 2)
+                loc = (dist * math.cos(angle), dist * math.sin(angle), 0) + agent_loc
```
Jialn: Can we use this style for all case including also_curriculum_target_angle=False by setting _random_angle=2*np.pi ?
given the same "range", distribution in polar coordinate and cartesian coordinate is different.

pygazebo/pygazebo.cc
```diff
@@ -394,7 +394,7 @@ class World {
   std::string ModelListInfo() {
     std::stringstream ss;
     for (auto model : world_->Models()) {
-      ss << "Model: " << model->GetName() << std::endl;
+      ss << "Model: " << "\"" << model->GetName() << "\"" << std::endl;
```
emailweixu: You can use '"' instead of "\""

python/social_bot/teacher.py
```diff
@@ -341,7 +346,7 @@ def teach(self, agent_sentence):
                 g = self._task_groups.pop(active_group_id)
                 self._task_groups.insert(0, g)
             return_action = TeacherAction(final_reward, final_sentence, done)
-        logging.debug("Teacher Reward: %f, Sentence: %s, Done: %d",
-                      return_action.reward, return_action.sentence,
-                      return_action.done)
+        # logging.debug("Teacher Reward: %f, Sentence: %s, Done: %d",
+        #               return_action.reward, return_action.sentence,
+        #               return_action.done)
```
emailweixu: remove or uncomment

python/social_bot/tasks.py
emailweixu: @JiaLn is right. Here the maximum distance is range. It's range/sqrt(2) at line 370. And line 370 put higher probability for large distance.

python/social_bot/tasks.py
Jialn: perhaps name to curriculum_action_scale is better

python/social_bot/envs/play_ground.py
emailweixu: Why?

python/social_bot/envs/play_ground.py
```diff
@@ -235,14 +235,18 @@ def step(self, action):
             If with_language, it is a dictionary with key 'data' and 'sentence'
             If not with_language, it is a numpy.array or image for observation
         """
-        if self._with_language:
+        if False and self._with_language:
```
emailweixu: why False?

python/social_bot/envs/play_ground.py
le-horizon: during training, when computing logProb, it complains TypeError: Value passed to parameter 'x' has DataType uint32 not in list of allowed values...

Maybe somewhere we or TF is translating it into uint32.  I haven't got a chance to look deeper into it yet, and just picked a simpler fix to make our agent mute..

python/social_bot/envs/play_ground.py
le-horizon: same as above

python/social_bot/teacher.py
le-horizon: default get_current_task returns a python generator function, instead of the task object.  Maybe we should rename the default one as well?

python/social_bot/envs/play_ground.py
Jialn: We can set with_language to False through gin file to mute the agent.

python/social_bot/tasks.py
Jialn: I think target random_range is fine because the agent have the target pos in observation. It's just a new range of an existing feature in observation. If random_range changes, target position in the next state s' still depends on last states and actions, does not depends on random_range. 

But if action_range changes, it's a different case. The next state s' depends on last states, actions and action_range.

For the flipping problem, either we can set the effort_limit in model file a little bit lower, or add a slightly action_cost term. 

Another way is to set the joint using a PID controller rather than force controlling directly. It's vey simple, just by editing the agent config file in "agent.cfg" to enable this. I would suggest this way and I already set Youbot and iCub joints to PID. The action is the target speed of PID controller input. Actually most real robots, robot arms or quadcopters using sub-level PID controller, either velocity or position way.

Force controlling is dangers because F/m is the acceleration, given enough time*action for one side of wheels the velocity could be very high, even action range is smaller.

python/social_bot/tasks.py
Jialn: I prefer to polar cord.

python/social_bot/teacher.py
emailweixu: Seems that this function is not used anywhere

python/social_bot/envs/play_ground.py
emailweixu: There are two fields for action, one is `control`, the other is `sentence`. The current ActionDistributionNetwork probably cannot handle it. Perhaps we should add a config like `with_agent_language`

le-horizon:
@emailweixu I rebased the change on new master, could you take a quick look?

Thanks,
Le

Issue #102, opened 2019-11-13
mesh and texture files of some models is not used


Jialn:
https://github.com/HorizonRobotics/SocialRobot/pull/97
PR97 added all used gazebo model files into repository, but mesh and texture files of the model that are renamed is not used. The link in the sdf is still the original one. Currently we still need to downloading model files to ~/.gazebo/models to get the texture correctly.

example: here is part of the codes pr2_noplugin.sdf

```
      <collision name="torso_lift_link_geom">
        <geometry>
          <mesh>
            <uri>model://pr2/meshes/torso_v0/torso_lift_L.stl</uri>
          </mesh>
        </geometry>
```
Either we should change pr2 in the uri to pr2_noplugin, or remove the mesh and texture file and add originlal pr2/pioneer/youbot to the repo.

le-horizon:
Good catch Jiangtao!  I guess changing URI to pr2_noplugin is the easiest?  Thanks!

Jialn:
Fxied

PR #103, opened 2019-11-14
Add CI test using Github Actions

Jialn:
Added build and code checking for playground_test. 

The script installs gazebo dependency every time. To avoid this I also build a docker image https://hub.docker.com/layers/jialn/socialbot/latest/images/sha256-6db01ed73b6516995a4dbf971d4228a53baf1f0c00a91213fea1be468c6defca

But it seems that installing gazebo dependency (need to get some 200~400MB from Azure internal sources) is fast enough, and maybe faster than pulling image from docker-hub (~800MB). The setup and installing progress take about 2.5 minutes.


.github/workflows/python_test.yml
```diff
@@ -0,0 +1,40 @@
...
+        mkdir build
+        cd build
+        cmake ..
+        make -j
+        cd ..
+        pip3 install -e .
+    - name: Lint with flake8
+      run: |
+        pip install flake8
+        # distable static check for now
```
emailweixu: distable => disable

.github/workflows/python_test.yml
```diff
@@ -0,0 +1,40 @@
...
+    steps:
+    - uses: actions/checkout@v1
+    - name: Set up Python 3.6
+      uses: actions/setup-python@v1
+      with:
+        python-version: 3.6
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip3 install -e .
```
emailweixu: This is a duplicate of line 29.

.github/workflows/python_test.yml
```diff
@@ -0,0 +1,40 @@
...
+        pip3 install absl-py pytest Pillow
```
emailweixu: should add these dependencies into setup.py instead.

Jialn: Done

.github/workflows/python_test.yml
Jialn: Fixed

.github/workflows/python_test.yml
Jialn: the first one is needed for cmake and compiling, another one to setup pygazebo after compiling.

.github/workflows/python_test.yml
```diff
@@ -0,0 +1,38 @@
...
+        make -j
+        cd ..
+        pip3 install -e .
+    - name: Lint with flake8
+      run: |
+        pip install flake8
+        # stop the build if there are Python syntax errors or undefined names
+        flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
+        # exit-zero treats all errors as warnings. The GitHub editor is 127 chars wide
+        flake8 ./python --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
```
emailweixu: It's better to follow https://github.com/HorizonRobotics/alf/blob/master/.ci-cd/build.sh to use the same style check to be consistent with ALF style.

.github/workflows/python_test.yml
```diff
@@ -0,0 +1,40 @@
...
+      run: |
+        pip install flake8
+        # distable static check for now
+        # stop the build if there are Python syntax errors or undefined names
+        # flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
+        # exit-zero treats all errors as warnings. The GitHub editor is 127 chars wide
+        # flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
+    - name: Test with pytest
+      run: |
+        GAZEBO_MODEL_PATH=./python/social_bot/models xvfb-run python python/social_bot/envs/play_ground_test.py
```
witwolf: Is it possible to run all testes by `GAZEBO_MODEL_PATH=./python/social_bot/models xvfb-run python -m unittest discover -p "*_test.py" -v` ?

.github/workflows/python_test.yml
Jialn: Added the the same style check using pre-commit. 
This flake8 checking can be kept, could do static checking for syntax errors or undefined names in corner case that tests does not cover, also it can report warnings like unused import, variables assigned to but never used. max-line-length checking can be ignored since pre-commit already checked that.

.github/workflows/python_test.yml
```diff
@@ -0,0 +1,48 @@
...
+        pip3 install pre-commit
+        sudo apt install clang-format-5.0
+        pre-commit install
+        if ! pre-commit run -a ; then
+          git diff
+          exit 0
+        fi
+    - name: Test with pytest
+      run: |
+        cd ./python/social_bot/envs
```
Jialn: "unittest discover" need to run in this path. Let's only run the tests under env for now.

Jialn:
https://github.com/HorizonRobotics/SocialRobot/issues/102 This issue generates lots of warnings about textures during the tests. I've removed unused texture and stl files in our customized models, and added the original models to our repo because of the texture dependency.

PR #106, opened 2019-11-19
add flags for not moving the ball after the goal is achieved and not considering the angle for achieving the goal also add a list for additional objects

ruizhaogit:
Hi guys,

In this pull request, I add two flags for not moving the ball after the goal is achieved and not considering the angle for achieving the goal also add a list for additional objects, including the gazebo model for a blue ball.

Best,
Rui

### DIFF
--- a/python/social_bot/models/blue/model.config
+++ b/python/social_bot/models/blue/model.config
@@ -1,0 +1,10 @@
+<?xml version="1.0"?>
+<model>
+  <name>blue</name>
+  <version>1.0</version>
+  <sdf version="1.5">model.sdf</sdf>
+
+  <description>
+    A blue ball.
+  </description>
+</model>
--- a/python/social_bot/tasks.py
+++ b/python/social_bot/tasks.py
@@ -352,8 +352,8 @@ class GoalTask(Task):
 
         for name in self._additional_observation_list:
             obj = self._world.get_model(name)
-            obj_pose = np.array(obj.get_pose()[0]).flatten() 
-            pose = np.concatenate((pose, obj_pose), axis=0)
+            obj_pos = np.array(obj.get_pose()[0]).flatten() 
+            pose = np.concatenate((pose, obj_pos), axis=0)
             
         return pose
 
--- a/python/social_bot/tasks.py
+++ b/python/social_bot/tasks.py
@@ -123,7 +123,10 @@ class GoalTask(Task):
                  reward_thresh_to_increase_range=0.4,
                  percent_full_range_in_curriculum=0.1,
                  max_reward_q_length=100,
-                 reward_weight=1.0):
+                 reward_weight=1.0,
+                 move_goal_during_episode=True,
+                 success_without_angle_requirement=False,
+                 additional_observation_list=[]):
         """
         Args:
             env (gym.Env): an instance of Environment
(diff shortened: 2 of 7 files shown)


python/social_bot/tasks.py
```diff
@@ -338,7 +348,14 @@ def task_specific_observation(self, agent):
             observation besides self states, for the non-image case
         """
         goal = self._world.get_model(self._goal_name)
-        return np.array(goal.get_pose()[0]).flatten()
+        pose = np.array(goal.get_pose()[0]).flatten()
+
+        for name in self._additional_observation_list:
+            obj = self._world.get_model(name)
+            obj_pose = np.array(obj.get_pose()[0]).flatten()
```
Jialn: we can the name "obj_pos = np.array(obj.get_pose()[0]).flatten()" here because pose includes rotation, and pos is only the position part.

python/social_bot/tasks.py
```diff
@@ -166,6 +172,9 @@ def __init__(self,
         self._random_goal = random_goal
         self._distraction_list = distraction_list
         self._object_list = distraction_list
+        self._move_goal_during_episode = move_goal_during_episode
+        self._success_without_angle_requirement = success_without_angle_requirement
```
Jialn: seems self._success_without_angle_requirement is not used.

python/social_bot/tasks.py
```diff
@@ -259,7 +268,8 @@ def run(self):
                 agent_sentence = yield TeacherAction(
                     reward=reward, sentence="well done", done=False)
                 steps_since_last_reward = 0
-                self._move_goal(goal, loc)
+                if self._move_goal_during_episode:
```
Jialn: if not move_goal_during_episode, should we send a "done" here to end the episode?

python/social_bot/tasks.py
le-horizon: @ruizhaogit, probably missed that part of the change?

python/social_bot/tasks.py
le-horizon: Rui is training pioneer to control the first goal (ball) to hit the other goal (blue_ball), with mutual information reward to encourage two balls to move together.  So reaching target shouldn't end episode.

ruizhaogit: Done is not needed here because reaching target shouldn't end episode.

python/social_bot/tasks.py
ruizhaogit: It is now renamed to "obj_pos = np.array(obj.get_pose()[0]).flatten()" in the new commits.

python/social_bot/tasks.py
ruizhaogit: The missing part is added in new commits.

python/social_bot/models/blue/model.config
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0"?>
+<model>
+  <name>blue</name>
```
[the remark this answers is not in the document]
ruizhaogit: Renamed to blue_ball also included your change to avoid name collision in the new commits.

python/social_bot/tasks.py
```diff
@@ -123,7 +123,10 @@ def __init__(self,
                  reward_thresh_to_increase_range=0.4,
                  percent_full_range_in_curriculum=0.1,
                  max_reward_q_length=100,
-                 reward_weight=1.0):
+                 reward_weight=1.0,
+                 move_goal_during_episode=True,
+                 success_without_angle_requirement=False,
```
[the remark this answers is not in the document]
ruizhaogit: It is renamed to "success_with_angle_requirement=True" in the new commits.

PR #108, opened 2019-11-21
Add Kuka LWR4 robot and Reaching3D task

Jialn:
Changes:
 1. Add Kuka LWR4+ robot
 2. Add Reaching3D task and a SAC training example

This is a fast-converging task, so it's good to serve as a environment to testing our algorithms.

It takes 6-7 minutes to solve the task with distance based reward shaping, and takes 1h15 minutes if without reward shaping.

reaching3d_sac_kukalwr4_rewardshaping

reaching3d_sac_kukalwr4_sparsereward

I also accidentally(by a bug) did a 2D version experiment, which takes 4-5 minutes to solve the task with reward shaping, and takes 30 minutes if without reward shaping.

reaching3d_sac_kukalwr4_rewardshaping

reaching3d_sac_kukalwr4_sparsereward


examples/reaching3d_sac_kukalwr4.gin
```diff
@@ -0,0 +1,23 @@
...
+create_environment.env_name="SocialBot-PlayGround-v0"
+create_environment.num_parallel_environments=12
+create_environment.env_load_fn=@suite_socialbot.load
+
+PlayGround.agent_type="kuka_lwr_4plus"
+PlayGround.tasks=[@Reaching3D]
+Reaching3D.reward_shaping=True
+PlayGround.step_time=0.05  # 20Hz step frequence
+PlayGround.max_steps=50  # 2.5s
+# make sure max_episode_steps of gym does not override PlayGround.max_steps
```
Haichao-Zhang: Do you mean max_episode_steps of gym should be larger than PlayGround.max_steps?

Jialn: Yes. So that the expected return of last step of an episode is correctly handled.

Jialn: Possible situations:

If max_episode_steps is reached and the env is reset, the potential discounted future reward will be take into account when update the value network. If not ended in purpose, most task should be like this by default . (max_episode_steps < max_steps)

If max_steps is reached with 'done' and the env is reset, the algorithm don't know env reaches time limits and treat this the same as done. All potential discounted reward in the future will be considered as zero. For some task like goal task and reaching task we use this way. (max_episode_steps > max_steps with done for each ending case)

If max_steps is reached without 'done' and the env is reset, the algorithm don't know env is reset, the initial state of next episode will be considered as the next state of the last step. This should never happen. (max_episode_steps > max_steps without done)

python/social_bot/tasks.py
```diff
@@ -683,3 +679,84 @@ def _move_ball(self, ball, goal_loc):
             if np.linalg.norm(loc - goal_loc) > self._success_distance_thresh:
                 break
         ball.set_pose((loc, (0, 0, 0)))
+
+
+@gin.configurable
+class Reaching3D(Task):
+    """
+    A task to reaching a random 3D position (r, theta, phi) with its finger tip or gripper
```
emailweixu: reaching => reach

python/social_bot/tasks.py
```diff
@@ -683,3 +679,84 @@ def _move_ball(self, ball, goal_loc):
...
+    An optional distance based reward shaping can be used.
+    This task is only compatiable for Agent kuka_lwr_4plus.
```
emailweixu: comptiable => compatible

python/social_bot/tasks.py
```diff
@@ -54,8 +54,8 @@ def __init__(self, env, max_steps=200, reward_weight=1.0):
 
     @abstractmethod
     def run(self):
-        """
-        run() use yield to generate TeacherAction
+        """ run() use yield to generate TeacherAction
```
emailweixu: Returns: A generator of TeacherAction

Jialn:
Comments fixed. Also updated README.

Issue #109, opened 2019-11-26
Clean up gazebo error/warning messages


emailweixu:
There are a lot of gazebo error/warning messages. They are confusing the new users. We should find ways to avoid them:
```
[Err] [ConnectionManager.cc:472] Data was empty, try again
[Err] [Connection.hh:268] Header is empty
[Err] [LogRecord.cc:158] LogRecord has not been initialized.
Warning [parser.cc:531] Converting a deprecated SDF source[data-string].
```

Jialn:
add another one:

`[Wrn] [Event.cc:61] Warning: Deleting a connection right after creation. Make sure to save the ConnectionPtr from a Connect call`

Jialn:
"LogRecord has not been initialized" can be fixed by add initialization to LogRecord. Will fix it soon.

Jialn:
"Warning [parser.cc:531] Converting a deprecated SDF source[data-string]" 

The warning is generated at "sdf.SetFromString(sdfString)" at line 462 of pygazebo.cc. It is printed at line 361 at "/gazebo/sdf/interface/parser.cc". 
https://bitbucket.org/osrf/gazebo/src/eae3d1b7adce37b2838501a25e42d36966dce298/gazebo/sdf/interface/parser.cc?at=default

Any insetting model or loading world using a sdf string will throw this warning, and it seems can not be disabled. As a woking around, perhaps we can fix this by writing the string to a file and load world or model by the file.

Jialn:
```
[Err] [ConnectionManager.cc:472] Data was empty, try again
[Err] [Connection.hh:268] Header is empty
```
Where does this come out? I haven't see them on my PC.

emailweixu:
> ```
> [Err] [ConnectionManager.cc:472] Data was empty, try again
> [Err] [Connection.hh:268] Header is empty
> ```
> 
> Where does this come out? I haven't see them on my ubuntu PC.

I have a lot of these messages during the training.

Jialn:
```
[Err] [LogRecord.cc:158] LogRecord has not been initialized.
Warning [parser.cc:531] Converting a deprecated SDF source[data-string].
```
These 2 have been sovled in https://github.com/HorizonRobotics/SocialRobot/pull/110

Jialn:
```
[Err] [ConnectionManager.cc:472] Data was empty, try again
[Err] [Connection.hh:268] Header is empty
[...]

PR #110, opened 2019-11-26
Fix Sdf Version Warning and LogRecord Er

Jialn:
Fix some warning of https://github.com/HorizonRobotics/SocialRobot/issues/109

Changes:
Fix Sdf Version Warning
Fix LogRecord Er
Also fixed the test of play ground because of the new option with_agent_language.

### DIFF
--- a/pygazebo/pygazebo.cc
+++ b/pygazebo/pygazebo.cc
@@ -485,8 +485,9 @@ void Initialize(const std::vector<std::string>& args,
     {
       gazebo::util::LogRecordParams params;
       params.period = 1e300;  // In fact, we don't need to do logging.
-      // gazebo::util::LogRecord::Instance()->Init("pygazebo");
+      gazebo::util::LogRecord::Instance()->Init("pygazebo");
       gazebo::util::LogRecord::Instance()->Start(params);
+      gazebo::util::LogRecord::Instance()->Stop();
     }
   }
   gazebo_initialized = true;
--- a/python/social_bot/envs/play_ground.py
+++ b/python/social_bot/envs/play_ground.py
@@ -125,7 +125,7 @@ class PlayGround(GazeboEnvBase):
         self._action_cost = action_cost
         self._with_language = with_language
         self._seq_length = vocab_sequence_length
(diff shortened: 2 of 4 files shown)


pygazebo/pygazebo.cc
```diff
@@ -485,8 +485,9 @@ void Initialize(const std::vector<std::string>& args,
     {
       gazebo::util::LogRecordParams params;
       params.period = 1e300;  // In fact, we don't need to do logging.
-      // gazebo::util::LogRecord::Instance()->Init("pygazebo");
+      gazebo::util::LogRecord::Instance()->Init("pygazebo");
       gazebo::util::LogRecord::Instance()->Start(params);
+      gazebo::util::LogRecord::Instance()->Stop();
```
Jialn: We should stop it immediately, otherwise more warnings and errors will occur during the progress of LogRecord.

python/social_bot/envs/play_ground.py
```diff
@@ -125,7 +125,7 @@ def __init__(self,
         self._action_cost = action_cost
         self._with_language = with_language
         self._seq_length = vocab_sequence_length
-        self._with_agent_language = with_agent_language
+        self._with_agent_language = with_language and with_agent_language
```
Jialn: it seems we should add this logic to prevent the case with_language is false and with_agent_language is true.

Perhaps we can change them to something like with_language_in_observation and  with_language_in_action

le-horizon: Thanks for the fix Jiangtao, LG.  If we have more than 2 sources of language then at that point we can do _in_obs, _in_action etc.?

python/social_bot/envs/play_ground.py
```diff
@@ -260,11 +260,13 @@ def main():
     import matplotlib.pyplot as plt
```
le-horizon: There is a separate play_ground_test.py, maybe we can remove this test?

Jialn: It's till useful when we want render, test specific case, or run for a long time to test memory growth or something. Let's keep it.

PR #112, opened 2019-11-28
Add world.step() to reset()

Jialn:
Sorry that "self._world.step()" has been removed accidentally in PR https://github.com/HorizonRobotics/SocialRobot/pull/104 when remove the randomness by random action taken at the start of each episode. It is needed to update the observation after reset.

Also fixed a problem forgot to update the parameters of "get_angle_to_target()" function being updated in PR https://github.com/HorizonRobotics/SocialRobot/pull/104.

### DIFF
--- a/python/social_bot/envs/play_ground.py
+++ b/python/social_bot/envs/play_ground.py
@@ -194,6 +194,7 @@ class PlayGround(GazeboEnvBase):
         self._steps_in_this_episode = 0
         self._world.reset()
         self._teacher.reset(self._agent, self._world)
+        self._world.step(self._sub_steps)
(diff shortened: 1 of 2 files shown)


python/social_bot/envs/play_ground.py
```diff
@@ -194,6 +194,7 @@ def reset(self):
         self._steps_in_this_episode = 0
         self._world.reset()
         self._teacher.reset(self._agent, self._world)
+        self._world.step(self._sub_steps)
```
emailweixu: This should be after line 199 because teacher may make some change to the environment (e.g. randomly move objects)

PR #116, opened 2019-12-16
Update keyboard demo generation and add action wrapper

Jialn:
The keyboard control require and it is not convenient to use keys to control the arm joints. So I change that to using PyUserInput to grab mouse and keyboard input so that we don't need to focus on the terminal window, and easier arm joints controlling.

I've also add an optional action wrapper. The human demonstrations is not the same as raw joints. By this action wrapper, we can transform the action to the same space as demonstrations. And sometimes primitive joints is not wanted, e.g., has redundant dimensions or offset, like extra wheel joints, or 2 joints of finger on a gripper which can be combined to one single action dimension. In these cases, the wrapper can also be used to simplify the actions.

This PR also includes some changes to youbot model, like adding camera and better joints range and initial position.

Changes:
- Arm joints demo generation by mouse
- Use PyUserInput to grab mouse and keyboard input so that we don't need to focus on the terminal window
- Add an optional action wrapper
- Youbot model adjustment: 
     - Add camera;
     - Back off the base of robot arm a little bit to spare the space for camera; 
     - Better joints range and initial position: The original model range always starts from 0, eg., (0, 2*pi); but the commonly used range (where gripper is in front of the robot) is not sure, might be like 150-210 degree or 0 ~ 30 and 330-360 degree.  That is both un-convenient for human control and RL agent. The policy net work has to learn the offset, and sometimes has to overcome the uncontinous jump between 6.28 and 0. This cause some wired behavior when training the agent. I've center the commonly used range to 0 and make the joints range continuous at the center using action wrapper.

Camera position:
image

Now we can generate demonstrations using the keyboard key "W, A, S, D", "E" and mouse by watching the 128x128 image easily. This can be used to demonstrate many versatile teacher tasks.
pick_place_img_2


python/social_bot/tasks.py
```diff
@@ -906,9 +906,9 @@ class PickAndPlace(Task):
     def __init__(self,
                  env,
                  max_steps,
-                 object_random_range=0.5,
-                 place_to_random_range=0.5,
-                 min_distance=0.25,
+                 object_random_range=0.6,
+                 place_to_random_range=0.6,
+                 min_distance=0.3,
```
Jialn: Sometimes cube is moved into the body of youbot if min_distance is 0.25. 0.3 is more safer.

python/social_bot/keybo_control.py
```diff
@@ -203,10 +192,6 @@ def main():
         if done:
             env.reset()
             keybo.reset()
-            step_per_sec = step_cnt / (time.time() - last_done_time)
-            logging.info("step per second: " + str(step_per_sec))
```
Jialn: fps is limited by "real_time_update_rate", there is no need to count fps here.

python/social_bot/gazebo_agent.py
```diff
@@ -341,3 +352,74 @@ def get_contacts(self, contacts_sensor, contact_collision):
...
+        "arm_joint_yaw", "arm_joint_pitch", "arm_joint_pitch_2", "palm_joint",
+        "gripper_finger_joint", "wheel_speed", "wheel_turning"
+    ]
+    The primitive actions (the joints) please refer to social_bot/models/agent_cfg.json.
+    """
+
+    def get_actions_dim(self):
+        """ Get the dimension of the new action
+        """
+        return 7
```
Haichao-Zhang: Might be better to change the hard-coded number of actions by 1) maintaining a list of actions and 2) calculating the length of the list as actions_dim ?

python/social_bot/gazebo_agent.py
```diff
@@ -341,3 +352,74 @@ def get_contacts(self, contacts_sensor, contact_collision):
...
+
+    def warp_actions(self, action):
```
Haichao-Zhang: warp_actions => wrap_actions ?

python/social_bot/gazebo_agent.py
```diff
@@ -341,3 +352,74 @@ def get_contacts(self, contacts_sensor, contact_collision):
...
+    dimensions or offsets. By the action wrapper, we can transform the action to more
+    efficency one.
+    """
+
+    def get_actions_dim(self):
+        """ Get the dimension of the new action space
+        """
+        raise NotImplementedError("get_actions_dim not implemented!")
+
+    def warp_actions(self, action):
```
Haichao-Zhang: wrap_actions?

python/social_bot/gazebo_agent.py
```diff
@@ -341,3 +352,74 @@ def get_contacts(self, contacts_sensor, contact_collision):
...
+        """ Wrap transformed actions to primitive actions.
+
+        Args:
+            action (nparray): the new action from policy network
+        Returns:
+            np.array, the primitive actions send to simulator
+        """
+        raise NotImplementedError("warp_actions not implemented!")
```
Haichao-Zhang: wrap_actions?

python/social_bot/keybo_control.py
```diff
@@ -120,38 +125,22 @@ def _to_diff_drive_action(self):
         return actions
 
     def _to_youbot_action(self):
-        wheel_joint_fl = self._speed + self._turning
-        wheel_joint_fr = self._speed - self._turning
+        """ Convert to the warpped youbot actions
```
Haichao-Zhang: wrapped

python/social_bot/gazebo_agent.py
```diff
@@ -122,6 +131,8 @@ def take_action(self, action):
         Args:
             the actions to be taken.
         """
+        if self._action_warpper is not None:
```
Haichao-Zhang: might beed to update accordingly if decided to change from warp to wrap

python/social_bot/gazebo_agent.py
Jialn: sorry, should be wrap...

python/social_bot/gazebo_agent.py
Jialn: Done

PR #117, opened 2019-12-18
Add simple internal states for goal task

le-horizon:
Dear reviewers,

This PR includes several changes, submitting for review before it gets too large.

1) added a mode of simple full world states as input to the agent.  The input is basically the goal and the distraction object's location in the agent's own viewpoint/coordinate-system.  Also includes a mode to limit the view angle of the agent.  This imitates perfect attention based on visual input, and is helpful in debugging system performance.

2) a few optimizations and fixes for the task:
a) place goal task objects within the boundary of the walls,
b) added distraction objects' states to observation by default,
c) allow using language input (without language, goal location is always the first in the input, followed by locations of the distractions, even when goal object changes e.g. from ball to car_wheel; with language, input locations are always ordered by the fixed order of objects in the world, and language specifies which of them is the goal),
d) using velocity (PID) control for pioneer by default, which cuts training time to half.

Thanks,
Le


python/social_bot/gazebo_agent.py
```diff
@@ -54,6 +58,13 @@ def __init__(self,
                 if None it will be set the same as agent_type
             config (dict): the configuarations for the agent
                 see `agent_cfg.jason` for details
+            use_pid_control (bool): whether use pid controller for agent.
+                Note: not all agents have pid control setup.  Check agent_cfg.json.
+            use_simple_full_states (bool): Use the simplest full states like
+                agent's distance and direction to goal
+            view_angle_limit (float): the radian angle to limit the agent's observation
```
emailweixu: better to use degree than radian. It's easier to specify

python/social_bot/gazebo_agent.py
```diff
@@ -180,10 +195,37 @@ def get_full_states_observation(self, teacher):
...
+            # task_specific_observation give the goal position.
+            yaw = agent_pose[5]
+            obs = []
+            while len(task_specific_ob) > 1:
+                x = task_specific_ob[0] - agent_pose[0]
+                y = task_specific_ob[1] - agent_pose[1]
+                # rotate -yaw
+                rotated_x = x * np.cos(-yaw) - y * np.sin(-yaw)
+                rotated_y = x * np.sin(-yaw) + y * np.cos(-yaw)
+                if self._view_angle_limit > 0.001:
```
emailweixu: why 0.001? 0 seems better.

python/social_bot/tasks.py
```diff
@@ -422,8 +427,12 @@ def _move_goal_impl(self, goal, agent_loc, agent_dir):
                    random.random() * range - range / 2, 0)
 
             self._initial_dist = np.linalg.norm(loc - agent_loc)
-            if self._initial_dist > self._success_distance_thresh:
+            if self._initial_dist > self._success_distance_thresh and (
+                attempts > 10000 or
+                (abs(loc[0]) < 5 and abs(loc[1]) < 5)  # within walls
```
emailweixu: There are a few places in the code where 5 is used. Should make it a variable.

python/social_bot/gazebo_agent.py
```diff
@@ -54,6 +58,13 @@ def __init__(self,
                 if None it will be set the same as agent_type
             config (dict): the configuarations for the agent
                 see `agent_cfg.jason` for details
+            use_pid_control (bool): whether use pid controller for agent.
```
Jialn: Perhaps we don't need this parameter to override default one in agent_cfg.json here. Use PID or not does not changes very often. Editing the "agent_cfg.json" will be fine.

python/social_bot/gazebo_agent.py
```diff
@@ -180,10 +195,37 @@ def get_full_states_observation(self, teacher):
         """
         task_specific_ob = teacher.get_task_specific_observation(self)
         agent_pose = np.array(self.get_pose()).flatten()
-        agent_vel = np.array(self.get_velocities()).flatten()
-        internal_states = self.get_internal_states()
-        obs = np.concatenate(
-            (task_specific_ob, agent_pose, agent_vel, internal_states), axis=0)
+        if self._use_simple_full_states:
```
Jialn: This seems to be dedicated to GoalTask. Maybe move these to GoalTask.specific_observation() ?

I thought agent_pose and agent_vel (including position and rotation) are commonly used before so I put them here. But it seems many task don't need all of them. Perhaps we can remove this "get_full_states_observation" function, directly use "get_task_specific_observation" instead. Let the tasks define all there observations for the non-image case. Both simpler and more flexible.

python/social_bot/gazebo_agent.py
```diff
@@ -180,10 +195,37 @@ def get_full_states_observation(self, teacher):
...
+            # assumes GoalTask and that the first 3 dims of the
+            # task_specific_observation give the goal position.
+            yaw = agent_pose[5]
+            obs = []
+            while len(task_specific_ob) > 1:
+                x = task_specific_ob[0] - agent_pose[0]
```
Jialn: Perhaps use angle directly would be a little bit simpler?
```
        object_direction = np.arctan2(object_pos[1] - agent_pos[1], object_pos[0] - agent_pos[0])
        view_angle = object_direction - yaw
        if view_angle > self._view_angle_limit or view_angle < -self._view_angle_limit:
                view_angle = 0
                obs.extend([view_angle, 0])
        else: 
                obs.extend([view_angle, 1])
```

python/social_bot/tasks.py
Jialn: We can remove the default grey wall in the play ground world file, and insert them according to maximal random range. 
The wall should be lower, and adjustable in length. We can add a customized model for that.

python/social_bot/gazebo_agent.py
le-horizon: Good point.  Let me remove this.

python/social_bot/gazebo_agent.py
le-horizon: Done

python/social_bot/gazebo_agent.py
le-horizon: I was being too careful in case float comparison isn't accurate.  Changed to 0.

python/social_bot/gazebo_agent.py
le-horizon: Good point.  I thought about that, but the problem with angle is it's discontinuous.  Coordinates also convey a sense of depth, that's why I'm using rotated_x and _y.

python/social_bot/tasks.py
le-horizon: Done

le-horizon: Very good point.  However, I'm hesitant to change the env as it makes old experiments ran on old env hard to compare with new runs on the new env. 
 I've added a TODO, and let's do this if absolutely needed?

python/social_bot/gazebo_agent.py
le-horizon: I think this code should still live here, as it's just changing from a global coordinate system to an egocentric one.  It could help tasks like pick and place as well?

Jialn: Different tasks have different task_specific_ob. It's hard to be compatible for all of them. Codes like 
```
x = task_specific_ob[0] - agent_pose[0]
task_specific_ob = task_specific_ob[3:]
```
will not work with other tasks since there might be touch sensors, rotations, angles, pos only contains x and y axis, etc.

So I still think that we should move them to GoalTask.specific_observation. For egocentric coordinate, we can add a common helper function to convert a global position to egocentric one given agent pose.

python/social_bot/tasks.py
Jialn: Not absolutely needed. Just a suggestion. I'll see if there is any elegant way to implement a wall model that is adjustable in length and height latter.

python/social_bot/gazebo_agent.py
le-horizon: I see, good catch Jiangtao!  I moved logic to taskspecific obs, and left only the flags in the GazeboAgent, which I think is a natural place for them to live.

Let me know how you like this way.

python/social_bot/tasks.py
le-horizon: Thanks Jiangtao.  It would be nice to have adjustable models, so we can control behavior using python.  We can also procedurally generate mazes etc..

python/social_bot/gazebo_agent.py
Jialn: Let's move the flag to task, and rename to something like use_egocentric_states. I think configure the observation egocentric or not in task is also very natural.

Jialn: agent_pose and agent_vel (including position and rotation) should be removed. That bring redundant or unused dimensions in some other task's observation. So remove "get_full_states_observation" directly use "get_task_specific_observation" would be better. Let the tasks define all there observations for the non-image case. Both simpler and more flexible.

This require to remove some redundant or unused dimensions for other task. Let me do this.

python/social_bot/gazebo_agent.py
Jialn: The code I wrote above is discontinuous...Sorry. We can first translate (x,y) to agent's coordinate, and then do the arctan2. So we have a range of arctan2 (-pi, pi).

If view_angle_limit < pi, the angle's range is (- view_angle_limit, view_angle_limit), there is no discontinuous problem.
If view_angle_limit >= pi, the range (-pi, pi). it is discontinuous when the object is behind the agent by 180 degree. But that's fine because the optimal action for the states is actually discontinuous: near to -pi, the agent should turn left, near to pi, the agent should turn right. This is a more "actionable" representation than coordinates.

python/social_bot/gazebo_agent.py
```diff
@@ -180,10 +195,37 @@ def get_full_states_observation(self, teacher):
...
+                y = task_specific_ob[1] - agent_pose[1]
+                # rotate -yaw
+                rotated_x = x * np.cos(-yaw) - y * np.sin(-yaw)
+                rotated_y = x * np.sin(-yaw) + y * np.cos(-yaw)
+                if self._view_angle_limit > 0.001:
+                    dist = math.sqrt(rotated_x * rotated_x + rotated_y * rotated_y)
+                    rotated_x /= dist
+                    rotated_y /= dist
+                    magnitude = 1. / dist
```
Jialn: Why not use dist directly here? If agent get very closer to the goal, dist is very small, will magnitude be very large?

python/social_bot/gazebo_agent.py
le-horizon: Sounds good.  Thanks!

python/social_bot/gazebo_agent.py
le-horizon: Good point.  Dist is usually not very close to 0 due to success dist thresh.  The reason to use magnitude is to make it change monotonically when object goes out of view, in which case the magnitude becomes 0.  I couldn't find a natural value if we use dist.

python/social_bot/tasks.py
```diff
@@ -460,8 +480,13 @@ def _move_goal_impl(self, goal, agent_loc, agent_dir):
                        random.random() * range - range / 2, 0)
 
             self._initial_dist = np.linalg.norm(loc - agent_loc)
-            if self._initial_dist > self._success_distance_thresh:
+            if self._initial_dist > self._success_distance_thresh and (
+                attempts > 10000 or (
```
Jialn: In what case will the attempts > 10000 and still can not found a valid position? Maybe log.e() some information if attempts > 10000 ?

python/social_bot/tasks.py
```diff
@@ -499,18 +524,59 @@ def task_specific_observation(self, agent):
...
+                    obs.extend([rotated_x, rotated_y, magnitude])
+                else:
+                    obs.extend([rotated_x, rotated_y])
+                pose = pose[3:]
+            obs = np.array(obs)
+        else:
+            agent_vel = np.array(agent.get_velocities()).flatten()
+            joints_states = agent.get_internal_states()
+            obs = np.concatenate(
+                (pose, agent_pose, agent_vel, joints_states), axis=0)
```
Jialn: If use_egocentric_states, joints_states (the sin and cos of wheel joint angles and wheel speed) is not included. Is this on purpose?

le-horizon: Yes, I just haven't experimented it.  But good point.  Let me try and see if that improves things!

Jialn:
LGTM

le-horizon:
@emailweixu looks like I still need your approval to merge.

Could you take a look?

Thanks,
Le

Jialn:
the last commit fails
```
ERROR
128
test_pr2 (pr2_test.TestPr2) ... [Wrn] [IntrospectionManager.cc:115] Item [data://world/default/model/ground_plane?p=pose3d/world_pose] already registered
129
[Wrn] [IntrospectionManager.cc:115] Item [data://world/default/model/ground_plane?p=vector3d/world_linear_velocity] already registered
130
[Wrn] [IntrospectionManager.cc:115] Item [data://world/default/model/ground_plane?p=vector3d/world_angular_velocity] already registered
131
[Wrn] [IntrospectionManager.cc:115] Item [data://world/default/model/ground_plane?p=vector3d/world_linear_acceleration] already registered
132
[Wrn] [IntrospectionManager.cc:115] Item [data://world/default/model/ground_plane?p=vector3d/world_angular_acceleration] already registered
133
[Wrn] [IntrospectionManager.cc:115] Item [data://world/default/model/ground_plane/link/link?p=pose3d/world_pose] already registered
134
[Wrn] [IntrospectionManager.cc:115] Item [data://world/default/model/ground_plane/link/link?p=vector3d/world_linear_velocity] already registered
135
[Wrn] [IntrospectionManager.cc:115] Item [data://world/default/model/ground_plane/link/link?p=vector3d/world_angular_velocity] already registered
136
[Wrn] [IntrospectionManager.cc:115] Item [data://world/default/model/ground_plane/link/link?p=vector3d/world_linear_acceleration] already registered
137
[Wrn] [IntrospectionManager.cc:115] Item [data://world/default/model/ground_plane/link/link?p=vector3d/world_angular_acceleration] already registered
138
python3: /usr/include/boost/smart_ptr/shared_ptr.hpp:734: typename boost::detail::sp_member_access<T>::type boost::shared_ptr<T>::operator->() const [with T = gazebo::physics::Entity; typename boost::detail::sp_member_access<T>::type = gazebo::physics::Entity*]: Assertion `px != 0' failed.
139
Aborted (core dumped)
140
##[error]Process completed with exit code 134.
```

le-horizon:
Yes, looks like I can't run the test remotely.  I'll get to it coming Monday.
It's not so clear to me why it's failing..  The change shouldn't cause insertion of duplicate objects..

Jialn:
I've run it locally, seems the problem is that some locations (agent_loc/model_loc) is not converted to ndarray.
```
  File "/home/ubuntu/Srcs/SocialRobot/SocialRobot/python/social_bot/envs/play_ground.py", line 186, in __init__
    self.reset()
  File "/home/ubuntu/Srcs/SocialRobot/SocialRobot/python/social_bot/envs/play_ground.py", line 203, in reset
    teacher_action = self._teacher.teach("")
  File "/home/ubuntu/Srcs/SocialRobot/SocialRobot/python/social_bot/teacher.py", line 335, in teach
    teacher_action = g.teach(agent_sentence)
  File "/home/ubuntu/Srcs/SocialRobot/SocialRobot/python/social_bot/teacher.py", line 88, in teach
    teacher_action = task.send(agent_sentence)
  File "/home/ubuntu/Srcs/SocialRobot/SocialRobot/python/social_bot/tasks.py", line 354, in run
    self._move_goal(goal, loc, agent_dir)
  File "/home/ubuntu/Srcs/SocialRobot/SocialRobot/python/social_bot/tasks.py", line 464, in _move_goal
    avoid_locations=avoid_locations)
  File "/home/ubuntu/Srcs/SocialRobot/SocialRobot/python/social_bot/tasks.py", line 501, in _move_obj
    dist = np.linalg.norm(loc - avoid_loc)
TypeError: unsupported operand type(s) for -: 'tuple' and 'tuple'
  In call to configurable 'PlayGround' (<function PlayGround.__init__ at 0x7fab9b133730>)
```

le-horizon:
You are right Jiangtao!  I fixed the return value to be np array, and now tests are passing.

Jialn:
There are many warnings in "checking code style" in both the new added codes and existing codes like below. Perhaps run "pre-commit run -a" to auto-format them, or add the hook before commit?
```
@@ -459,13 +466,19 @@ class GoalTask(Task):
             for item, _ in distractions.items():
                 distraction = self._world.get_agent(item)
                 loc = self._move_obj(
-                    obj=distraction, agent_loc=agent_loc,
-                    agent_dir=agent_dir, is_goal=False,
+                    obj=distraction,
+                    agent_loc=agent_loc,
+                    agent_dir=agent_dir,
+                    is_goal=False,
                     avoid_locations=avoid_locations)
                 avoid_locations.append(loc)
 
-    def _move_obj(
-        self, obj, agent_loc, agent_dir, is_goal=True, avoid_locations=[]):
+    def _move_obj(self,
+                  obj,
+                  agent_loc,
+                  agent_dir,
+                  is_goal=True,
+                  avoid_locations=[]):
```

le-horizon:
Good point Jiangtao!  I have precommit hooks setup for alf, but not for socialrobot.  Is "pre-commit install" the way to go?

Btw, pushed the style fix.

python/social_bot/tasks.py
```diff
@@ -257,15 +268,23 @@ def __init__(self,
...
-        self._success_with_angle_requirement = success_with_angle_requirement
-        self._additional_observation_list = additional_observation_list
         if goal_name and goal_name not in distraction_list:
             self._object_list.append(goal_name)
         self._goals = self._object_list
-        self._pos_list = list(itertools.product(range(-5, 5), range(-5, 5)))
+        self._move_goal_during_episode = move_goal_during_episode
+        self._success_with_angle_requirement = success_with_angle_requirement
+        if not additional_observation_list:
+            additional_observation_list = self._object_list
```
emailweixu: @le-horizon Why did you make this change? This change the observation. Several tasks if alf cannot run because of this change (https://github.com/HorizonRobotics/alf/pull/334/checks?check_run_id=445009810)

le-horizon: Hmm, this is so that full internal state input would include all objects in the world.
I'll check tmr how to fix the tests.  It's probably also a good idea to make alf tests run on socialbot PR's.

le-horizon: @emailweixu I tested without these two lines, still getting the same error as with the lines:

`  File "/home/lezhao/alf/alf/algorithms/misc_algorithm.py", line 163, in train_step
    add_batch()
  File "/home/lezhao/alf/alf/algorithms/misc_algorithm.py", line 160, in add_batch
    self._buffer.add_batch(feature_reshaped_tran)
  File "/home/lezhao/alf/alf/utils/data_buffer.py", line 94, in add_batch
    indices, tf.stop_gradient(bat[-n:])), self._buffer, batch)
  File "/home/lezhao/venv/lib/python3.5/site-packages/tensorflow_core/python/util/nest.py", line 535, in map_structure
    structure[0], [func(*x) for x in entries],
  File "/home/lezhao/venv/lib/python3.5/site-packages/tensorflow_core/python/util/nest.py", line 535, in <listcomp>
    structure[0], [func(*x) for x in entries],
  File "/home/lezhao/alf/alf/utils/data_buffer.py", line 94, in <lambda>
    indices, tf.stop_gradient(bat[-n:])), self._buffer, batch)
  File "/home/lezhao/venv/lib/python3.5/site-packages/tensorflow_core/python/ops/resource_variable_ops.py", line 1180, in scatter_nd_update
    name=name))
  File "/home/lezhao/venv/lib/python3.5/site-packages/tensorflow_core/python/ops/gen_state_ops.py", line 680, in resource_scatter_nd_update
    _six.raise_from(_core._status_to_exception(e.code, message), None)
  File "<string>", line 3, in raise_from
tensorflow.python.framework.errors_impl.InvalidArgumentError: Must have updates.shape = indices.shape[:batch_dim] + params_shape[slice_dim:], got updates.shape: [1,4,23], indices.shape: [1,1], params_shape: [100,3,23], slice_dim: 1, and batch_dim: 1 [Op:ResourceScatterNdUpdate]
  In call to configurable 'SyncOffPolicyDriver' (<function SyncOffPolicyDriver.__init__ at 0x7fb4abb8c2f0>)
  In call to configurable 'train_eval' (<function train_eval at 0x7fb4abb71d90>)
`

Is this what you saw?  Might be something else affecting it.

emailweixu:
Besides the change of GoalTask, some other changes of alf also cause the
problem. I fixed them in the updated PR.

PR #121, opened 2020-02-14
fix observation list

le-horizon:
Wei, let me know if this fixes the issue.  I'm still seeing the error locally, and maybe I don't have the up-to-date version of your branch.

Thanks,
Le


emailweixu:
You don't need to fix this anymore. I changed the code on the alf side to fit the the new observation. But we need to keep in mind the environments need to be backward compatible in the future.

le-horizon:
> You don't need to fix this anymore. I changed the code on the alf side to fit the the new observation. But we need to keep in mind the environments need to be backward compatible in the future.

Got it.  Will be more careful in the future.

PR #126, opened 2020-07-06
add 'success' field to TeacherAction

hnyu:
Currently only use this flag in PlayGround with GoalTask/PickAndPlace/KickBallTask. Other tasks will get a default value of False, which is compatible since we never had this info before.

Also add an option "sparse_reward" for KickBallTask.

### DIFF
--- a/.github/workflows/python_test.yml
+++ b/.github/workflows/python_test.yml
@@ -9,10 +9,10 @@ jobs:
 
     steps:
     - uses: actions/checkout@v1
-    - name: Set up Python 3.6
+    - name: Set up Python 3.7
       uses: actions/setup-python@v1
       with:
-        python-version: 3.6
+        python-version: 3.7
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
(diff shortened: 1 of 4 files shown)


.github/workflows/python_test.yml
```diff
@@ -45,4 +45,4 @@ jobs:
     - name: Test with pytest
       run: |
         cd ./python/social_bot/envs
-        GAZEBO_MODEL_PATH=../models xvfb-run python3 -m unittest discover -p "*_test.py" -v
+        GAZEBO_MODEL_PATH=../models PYTHONPATH=../:$PYTHONPATH xvfb-run python3 -m unittest discover -p "*_test.py" -v
```
emailweixu, line 48: Perhaps it's better to change the previous line to "cd ./python/social_bot"?

python/social_bot/tasks.py
```diff
@@ -388,7 +388,10 @@ def run(self):
                 self._push_reward_queue(max(reward, 0))
                 logging.debug("yielding reward: " + str(reward))
                 agent_sentence = yield TeacherAction(
-                    reward=reward, sentence="well done", done=False)
+                    reward=reward, sentence="well done",
+                    done=not (self._move_goal_during_episode
```
emailweixu, line 392: should keep compatibility with previous task. Add another flag end_episode_after_successs with default value False.

PR #127, opened 2020-07-08
make kickball task more realistic

hnyu:
The current implementation treats the entire gate as a point mass and the success condition is that the distance between the ball and the gate center is smaller than a threshold. This is unrealistic because normally one would assume that as long as the ball

1) passes the gate line, and
2) is between the two gate posts

Then it's a goal. 

So this PR changes the success condition to the above. As I checked, no existing example or the readme will be affected by this PR. Note that the existing reward shaping that's based on the distance between gate center and ball can still be used as an approximation.

Compared to the existing setup: The existing setup has a default value of 0.5 for the success distance threshold. Since the gate width is 2.0 then there will be certain areas near the two posts that won't be counted as successes even if the ball actually rolls into the gate.

### DIFF
--- a/python/social_bot/tasks.py
+++ b/python/social_bot/tasks.py
@@ -812,13 +812,13 @@ class ICubAuxiliaryTask(Task):
 @gin.configurable
 class KickingBallTask(Task):
     """
-    A simple task to kick a ball so that it rolls into the gate. An
+    A simple task to kick a ball so that it rolls into the goal. An
     optional reward shaping can be used to guide the agent run to the ball first:
-        Agent will receive 100 when succefully kick the ball into the gate.
+        Agent will receive 100 when succefully kick the ball into the goal.
         Agent will receive the speed of getting closer to the ball before touching the
             ball within 45 degrees of agent direction. The reward is trunked within
             parameter target_speed.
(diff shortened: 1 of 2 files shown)


python/social_bot/tasks.py
```diff
@@ -812,66 +812,69 @@ def task_specific_observation(self, agent):
...
         self._sparse_reward = sparse_reward
+        self._gate_distance = gate_distance
+        # By looking up the 'robocup_3Dsim_goal' model file:
+        self._gate_width = 2.1
+        self._gate_post_radius = 0.05
         self._env.insert_model(
             model="robocup_3Dsim_goal",
-            name="goal",
-            pose="-5.0 0 0 0 -0 3.14159265")
+            name="gate",
```
emailweixu, line 861: Actually, I believe goal is the correct way to name it.

hnyu: I can change it back. Just thought that 'goal' is easily confused with the general 'goal' meaning in RL.

hnyu: > Actually, I believe goal is the correct way to name it.

I can change it back. Just thought that 'goal' is easily confused with the general 'goal' meaning in RL.

PR #128, opened 2020-07-30
goal conditioned GoalTask and a few other small improvements

le-horizon:
the few improvements including
1. improving camera position to avoid penetration into other objects
2. allow non-ego centric full states observation in GoalTask
3. log info observation_desc

### DIFF
--- a/pygazebo/pygazebo.cc
+++ b/pygazebo/pygazebo.cc
@@ -507,6 +507,19 @@ void CloseWithoutModelbaseFini() {
   gazebo_sensor_initialized = false;
 }
 
+void CloseWithoutFini() {
+  // this causes segmentation fault when run.
+  gazebo::physics::stop_worlds();
+  gazebo::sensors::stop();
+  gazebo::util::LogRecord::Instance()->Stop();
+  gazebo::transport::stop();
+
+  gazebo::transport::fini();
+
+  gazebo_initialized = false;
+  gazebo_sensor_initialized = false;
+}
+
 void Close() {
   gazebo::shutdown();
   gazebo_initialized = false;
@@ -573,7 +586,7 @@ PYBIND11_MODULE(pygazebo, m) {
 
   m.def("close_without_model_base_fini",
         &CloseWithoutModelbaseFini,
-        "A customized close function without execute ModelbaseFini"
+        "A customized close function without execute ModelbaseFini. "
         "For some unknwon reason, ModelbaseFini() in the gazebo.shutdown() "
         "makes the"
         "process fail to exit when the environment is wrapped with process");
--- a/pygazebo/pygazebo.cc
+++ b/pygazebo/pygazebo.cc
@@ -507,19 +507,6 @@ void CloseWithoutModelbaseFini() {
   gazebo_sensor_initialized = false;
 }
 
-void CloseWithoutFini() {
-  // this causes segmentation fault when run.
-  gazebo::physics::stop_worlds();
-  gazebo::sensors::stop();
-  gazebo::util::LogRecord::Instance()->Stop();
-  gazebo::transport::stop();
-
-  gazebo::transport::fini();
-
-  gazebo_initialized = false;
-  gazebo_sensor_initialized = false;
-}
-
 void Close() {
   gazebo::shutdown();
   gazebo_initialized = false;
--- a/python/social_bot/envs/gazebo_base.py
+++ b/python/social_bot/envs/gazebo_base.py
@@ -138,6 +138,7 @@ class GazeboEnvBase(gym.Env):
         gazebo.close_without_model_base_fini()
         if self._rendering_process is not None:
             self._rendering_process.terminate()
+            self._rendering_process = None
 
     def insert_model(self, model, name=None, pose="0 0 0 0 0 0"):
         """
--- a/python/social_bot/envs/gazebo_base.py
+++ b/python/social_bot/envs/gazebo_base.py
@@ -134,6 +134,7 @@ class GazeboEnvBase(gym.Env):
     def close(self):
         super().close()
         gazebo.close_without_model_base_fini()
+        self.__del__()
 
     def insert_model(self, model, name=None, pose="0 0 0 0 0 0"):
         """
--- a/python/social_bot/envs/gazebo_base.py
+++ b/python/social_bot/envs/gazebo_base.py
@@ -134,7 +134,8 @@ class GazeboEnvBase(gym.Env):
     def close(self):
         super().close()
         gazebo.close_without_model_base_fini()
-        self.__del__()
+        if self._rendering_process is not None:
+            self._rendering_process.terminate()
 
     def insert_model(self, model, name=None, pose="0 0 0 0 0 0"):
         """
@@ -193,10 +194,6 @@ class GazeboEnvBase(gym.Env):
     def seed(self, seed=None):
         """Gym interface for setting random seed."""
         random.seed(seed)
-
-    def __del__(self):
-        if self._rendering_process is not None:
-            self._rendering_process.terminate()
 
 
 def _modify_world_xml(xml, modifications):
--- a/python/social_bot/envs/gazebo_base.py
+++ b/python/social_bot/envs/gazebo_base.py
@@ -82,6 +82,41 @@ class GazeboEnvBase(gym.Env):
         Args:
             mode (str): 'human' and 'rgb_array' is supported.
         """
+        if self._rendering_camera is None:
+            render_camera_sdf = """
+            <?xml version='1.0'?>
+            <sdf version ='1.6'>
+            <model name ='render_camera'>
+                <static>1</static>
+                <pose>%s</pose>
+                <link name="link">
+                    <sensor name="camera" type="camera">
+                        <camera>
+                        <horizontal_fov>0.95</horizontal_fov>
+                        <image>
+                            <width>640</width>
+                            <height>480</height>
+                        </image>
+                        <clip>
+                            <near>0.1</near>
+                            <far>100</far>
+                        </clip>
+                        </camera>
+                        <always_on>1</always_on>
+                        <update_rate>30</update_rate>
+                        <visualize>true</visualize>
+                    </sensor>
+                </link>
+            </model>
+            </sdf>
+            """
+            render_camera_sdf = render_camera_sdf % self._rendering_cam_pose
+            self._world.insertModelFromSdfString(render_camera_sdf)
+            time.sleep(0.2)
+            self._world.step(20)
+            self._rendering_camera = self._world.get_agent('render_camera')
+        image = self._rendering_camera.get_camera_observation(
+            "default::render_camera::link::camera")
         if mode == 'human':
             if self._rendering_process is None:
                 from subprocess import Popen
@@ -91,41 +126,6 @@ class GazeboEnvBase(gym.Env):
                 self._rendering_process = Popen(['gzclient'])
             return
         if mode == 'rgb_array':
-            if self._rendering_camera is None:
-                render_camera_sdf = """
-                <?xml version='1.0'?>
-                <sdf version ='1.6'>
-                <model name ='render_camera'>
-                    <static>1</static>
-                    <pose>%s</pose>
-                    <link name="link">
-                        <sensor name="camera" type="camera">
-                            <camera>
-                            <horizontal_fov>0.95</horizontal_fov>
-                            <image>
-                                <width>640</width>
-                                <height>480</height>
-                            </image>
-                            <clip>
-                                <near>0.1</near>
-                                <far>100</far>
-                            </clip>
-                            </camera>
-                            <always_on>1</always_on>
-                            <update_rate>30</update_rate>
-                            <visualize>true</visualize>
-                        </sensor>
-                    </link>
-                </model>
-                </sdf>
-                """
-                render_camera_sdf = render_camera_sdf % self._rendering_cam_pose
-                self._world.insertModelFromSdfString(render_camera_sdf)
-                time.sleep(0.2)
-                self._world.step(20)
-                self._rendering_camera = self._world.get_agent('render_camera')
-            image = self._rendering_camera.get_camera_observation(
-                "default::render_camera::link::camera")
             return np.array(image)
 
         raise NotImplementedError("rendering mode: " + mode +
--- a/python/social_bot/envs/gazebo_base.py
+++ b/python/social_bot/envs/gazebo_base.py
@@ -58,7 +58,9 @@ class GazeboEnvBase(gym.Env):
         if port is None:
             port = 0
         self._port = port
-        # to avoid different parallel simulation has the same randomness
+        # This avoids different parallel simulations having the same randomness.
+        # When calling from alf, alf.environments.utils.create_environment calls
+        # env.seed() afterwards, which is the real seed being used.  Not this one.
         random.seed(port)
         self._rendering_process = None
         self._rendering_camera = None
@@ -115,8 +117,6 @@ class GazeboEnvBase(gym.Env):
             time.sleep(0.2)
             self._world.step(20)
             self._rendering_camera = self._world.get_agent('render_camera')
-        image = self._rendering_camera.get_camera_observation(
-            "default::render_camera::link::camera")
         if mode == 'human':
             if self._rendering_process is None:
                 from subprocess import Popen
@@ -126,6 +126,8 @@ class GazeboEnvBase(gym.Env):
                 self._rendering_process = Popen(['gzclient'])
             return
         if mode == 'rgb_array':
+            image = self._rendering_camera.get_camera_observation(
+                "default::render_camera::link::camera")
             return np.array(image)
 
         raise NotImplementedError("rendering mode: " + mode +
--- a/python/social_bot/envs/play_ground.py
+++ b/python/social_bot/envs/play_ground.py
@@ -100,7 +100,9 @@ class PlayGround(GazeboEnvBase):
                 become an OrderedDict including ``achieved_goal``, ``desired_goal`` and
                 ``observation`` fields, instead of a flat np array.  ``achieved_goal`` will
                 contain agent's current 2-d (x, y) position and ``desired_goal`` will be
-                the goal object's 2-d position.  Reward and task termination remain unchanged.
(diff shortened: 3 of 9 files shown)


python/social_bot/envs/play_ground.py
```diff
@@ -93,6 +95,7 @@ def __init__(self,
             world_name (string): Select the world file, e.g., empty.world, play_ground.world,
                 grocery_ground.world
             tasks (list): a list of teacher.Task, e.g., GoalTask, KickingBallTask
+            goal_conditioned (bool): Turn on goal conditioned tasks.
```
emailweixu, line 98: comment what it the difference between goal conditioned task and non-goal-conditioned tasks, especially the observation.

le-horizon: Done.  PTAL @emailweixu 

Thanks,
Le

python/social_bot/tasks.py
```diff
@@ -597,19 +648,41 @@ def task_specific_observation(self, agent):
...
+                    [[0, rvx, rvy, vz], [1, a1, a2, a3],
+                     [2, agent_pose[0], agent_pose[1], agent_pose[2]],
+                     [3, agent_pose[3], agent_pose[4], agent_pose[5]]])
+                i = 4
+            else:
+                obj_array.extend([[0, rvx, rvy], [1, vz, a1], [2, a2, a3],
+                                  [3, agent_pose[0], agent_pose[1]],
+                                  [4, agent_pose[2], agent_pose[3]],
+                                  [5, agent_pose[4], agent_pose[5]]])
+                i = 6
```
emailweixu, line 673: This is very hacky.
A better way might be removing 662 - 673.  And in the loop starting from line 676, only fill obj_array, do not change obs. After that merge obs with obj_array.

python/social_bot/envs/gazebo_base.py
```diff
@@ -134,6 +134,7 @@ def render(self, mode='human'):
     def close(self):
         super().close()
         gazebo.close_without_model_base_fini()
+        self.__del__()
```
emailweixu, line 137: `__del__` is supposed to be called when the object is deleted. Strictly speaking, `close()` does not mean the object is deleted. So I suggest removing `__del__` and move the content here.

python/social_bot/envs/play_ground.py
```diff
@@ -186,6 +198,26 @@ def __init__(self,
...
+                desired_goal=goal_space,
+                achieved_goal=goal_space)
+
+    def _get_observation(self, teacher_sentence):
+        obs = self._agent.get_observation(self._teacher, teacher_sentence)
+        if self._goal_conditioned:
+            flat_obs = obs
+            obs = OrderedDict(observation=obs)
+            obs['achieved_goal'] = flat_obs[6:8]
+            obs['desired_goal'] = flat_obs[12:14]
```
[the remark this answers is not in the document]
le-horizon: I added two asserts above to protect against the cases where a different agent or a task other than the GoalTask is used.

Maybe it's best to have dict observations from the agent, so we can just get whatever fields we need here?
Let me know if you have a better idea.

python/social_bot/tasks.py
le-horizon: I've reverted this change.  The obj_array was added for adding an index to each object, so the objects can be selected using language instruction via e.g. GFT.  Removing as we are not pursuing that line of research at least for the moment.

le-horizon:
> where is observation_desc logged?

Currently I'm logging.info the obs desc.

python/social_bot/envs/play_ground.py
```diff
@@ -93,6 +95,14 @@ def __init__(self,
             world_name (string): Select the world file, e.g., empty.world, play_ground.world,
                 grocery_ground.world
             tasks (list): a list of teacher.Task, e.g., GoalTask, KickingBallTask
+            goal_conditioned (bool): Turn on goal conditioned tasks.  Currently only GoalTask
```
Jialn, line 98: What is purpose of this option? As I know, for goal task currently, full state observation already includes desired goal and agent's current position(achieved_goal).

python/social_bot/envs/play_ground.py
```diff
@@ -186,6 +200,28 @@ def __init__(self,
         self.reset()
         self.observation_space = self._agent.get_observation_space(
             self._teacher)
+        if self._goal_conditioned:
```
Jialn, line 203: Hi Le, 

I think It would be better to put "goal_conditioned" option and get_observation related code to GazeboAgent class in gazebo_agent.py, and keep play_ground.py clean.

Now we are procedding the details of observation and action in class Agent. For example, you can move the observation space building code to the function "GazeboAgent.get_observation_space()".

"play_ground.py" is not supposed to deal with details of ob, action and goal. This should be an option of the GoalTask. 


Add another OrderedDict support in agent.py ( get_observation_space and get_observation ) like:
```

        if self._image_with_internal_states or self._with_language:
            # observation is an OrderedDict with feild ....
            obs = self._create_observation_dict(teacher, sentence_raw)
        elif self._use_image_observation:  # observation is pure image
            obs = self.get_camera_observation()
        elif self._goal_conditioned:
            # observation is OrderedDict with feild ``achieved_goal``, ``desired_goal`` ....
            obs = self._create_gc_observation()
        else:  # observation is pure low-dimentional states
            obs = teacher.get_task_specific_observation(self)
        return obs

```
In self._create_gc_observation(), setup the observation with task_specific_observation, agent position and goal position.
In GoalTask, if self._goal_conditioned, do not include agent position and goal position.

-------

There are mainly 3 class for now: Task, Agent, and Playground:

- task.py contains the details implementation of task
- agent.py defines the interface of agent(like setup joints, build observation and action space, get image or execute action)
- play_ground.py put all of them together.

It might be better to make observation defined by each Task, because only task have all the infomation directly. But to reuse some common codes like building the observation/action space and image (with internal states) case, we defined the observation space in Agent.

python/social_bot/envs/play_ground.py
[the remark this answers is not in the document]
Jialn: Hi Le, Maybe you can move these codes to Agent class as I said above.

python/social_bot/tasks.py
```diff
@@ -598,18 +652,27 @@ def task_specific_observation(self, agent):
                 pose = np.concatenate((pose, obj_pos), axis=0)
 
         agent_pose = np.array(agent.get_pose()).flatten()
-        if self._use_egocentric_states:
+        if self._use_full_states or self._use_egocentric_states:
```
Jialn, line 655: what is the different between _use_full_states  or not?  task_specific_observation is suposed to be full states of the world. If essentially they are similar, remove the part in else below:
```
            agent_vel = np.array(agent.get_velocities()).flatten()
            joints_states = agent.get_internal_states()
            obs = np.concatenate((pose, agent_pose, agent_vel, joints_states),
                                 axis=0)
```

python/social_bot/tasks.py
```diff
@@ -238,10 +246,9 @@ def __init__(self,
             max_reward_q_length (int): how many recent rewards to consider when estimating agent accuracy.
             reward_weight (float): the weight of the reward, is used in multi-task case
             move_goal_during_episode (bool): if True, the goal will be moved during episode, when it has been achieved
-            end_episode_after_success (bool): if True, the episode will end once the goal is reached. A True value of this
-                flag will overwrite the effects of flags ``switch_goal_within_episode`` and ``move_goal_during_episode``.
             success_with_angle_requirement: if True then calculate the reward considering the angular requirement
             additional_observation_list: a list of additonal objects to be added
+            use_full_states (bool): For non-image observation, use full states of the world
```
Jialn, line 251: task_specific_observation() is supposed for non-image case and use full states of the world by default.  What is the difference use_full_states or not?

python/social_bot/tasks.py
```diff
@@ -598,18 +652,27 @@ def task_specific_observation(self, agent):
...
+                x = pose[0]
+                y = pose[1]
+                if self._use_egocentric_states:
+                    x = pose[0] - agent_pose[0]
+                    y = pose[1] - agent_pose[1]
+                    rotated_x, rotated_y = agent.get_egocentric_cord_2d(
+                        x, y, yaw)
+                else:
+                    rotated_x, rotated_y = x, y
                 if self._egocentric_perception_range > 0:
```
Jialn, line 676: Maybe self._use_egocentric_states and elf._egocentric_perception_range > 0 ? 
If _use_egocentric_states is False, rotated_x is not rotated, the angle caculation below will not be correct.

python/social_bot/envs/play_ground.py
le-horizon: This is mainly to return OrderedDict observation instead of flat array observation, as alf's hindsight relabeling depends on this format.  Modified comment to make it more clear.

python/social_bot/envs/play_ground.py
le-horizon: Good points JT!  Thanks for explaining in detail.   Moved to agent.

python/social_bot/tasks.py
le-horizon: full_states so far excludes agent's internal_states, but thinking of it, maybe we should include agent's internal states into aux_achieved (similar as agent's speed).

python/social_bot/tasks.py
le-horizon: good catch.  done.

le-horizon:
Thanks for the careful review @Jialn .  Done or answered all.

python/social_bot/tasks.py
Jialn: Internal states might be useless here for goal task. They are angles and velocity of two wheels. Pretty much another representation of position changes, velocity, yaw and yaw rate that are already in agent_vel. I think it's fine to remove this part.

le-horizon:
Merged in the changes.  I also added
1) multi dim reward
2) auxiliary achieved field
3) made human rendering consistent with rgb_array mode (in terms of pseudo randomness)

Please take another look.

Thanks,
Le

pygazebo/pygazebo.cc
```diff
@@ -507,6 +507,19 @@ void CloseWithoutModelbaseFini() {
   gazebo_sensor_initialized = false;
 }
 
+void CloseWithoutFini() {
```
Jialn, line 510: what is this used for?

python/social_bot/envs/gazebo_base.py
```diff
@@ -82,6 +84,39 @@ def render(self, mode='human'):
...
+                </link>
+            </model>
+            </sdf>
+            """
+            render_camera_sdf = render_camera_sdf % self._rendering_cam_pose
+            self._world.insertModelFromSdfString(render_camera_sdf)
+            time.sleep(0.2)
+            self._world.step(20)
+            self._rendering_camera = self._world.get_agent('render_camera')
         if mode == 'human':
```
Jialn, line 120: Seems that human rendering is still gazebo. Why moving the inserting process above could making human rendering consistent with rgb_array mode?

python/social_bot/gazebo_agent.py
```diff
@@ -170,6 +190,25 @@ def get_observation(self, teacher, sentence_raw="hello"):
             obs = self.get_camera_observation()
         else:  # observation is pure low-dimentional states
             obs = teacher.get_task_specific_observation(self)
+            if self._goal_conditioned:
+                flat_obs = obs
```
Jialn, line 194: maybe move into a new function for this part

python/social_bot/gazebo_agent.py
```diff
@@ -170,6 +190,25 @@ def get_observation(self, teacher, sentence_raw="hello"):
...
+                obs['achieved_goal'] = flat_obs[6:8]
+                obs['desired_goal'] = flat_obs[12:14]
+                if self._use_aux_achieved:
+                    obs['observation'] = flat_obs[14:]
+                    obs['aux_achieved'] = np.concatenate(
+                        (flat_obs[:6], flat_obs[8:12]), axis=0)
+                    if self._xy_only_aux:
+                        # 2: z-speed, 3, 4: angular velocity, 5: yaw-vel, 6, 7: x, y, 8: z, 9, 10, 11: pose
+                        obs['observation'] = np.concatenate(
+                            (flat_obs[2:5], flat_obs[8:11], flat_obs[14:]),
```
Jialn, line 206: Perhaps we should not use this style operation. I think a better way is to put them into GoalTask, since they are only for GoalTask so not supposed to be here, and GoalTask has the original data so we don't need to spilt numpy array:

```
if self._goal_conditioned:
        tasks = teacher.get_task_groups().get_tasks()
        assert_for_goal_conditioned_ob()
        tasks[0].get_goal_conditioned_ob()
```

And then add get_goal_conditioned_ob() in Goaltask, as well as the option self._use_aux_achieved and self._xy_only_aux to Task. We can also use it to get a sample observation to get feilds and shapes, so we don't have to use a fixed shape in "_construct_goal_condi_space()":


```
def construct_goal_condi_space:
        assert_for_goal_conditioned_ob
        get a sample observation from get_goal_conditioned_ob
        get feilds and shapes and set the space
```

```
def assert_for_goal_conditioned_ob:
        assert self.type == 'pioneer2dx_noplugin'
        assert not self._with_language
        assert not self._use_image_observation
        assert not self._image_with_internal_states
        tgs = teacher.get_task_groups()
        assert len(tgs) == 1
        tasks = tgs[0].get_tasks()
        assert len(tasks) == 1 and isinstance(tasks[0], GoalTask), str(tasks)

```

python/social_bot/tasks.py
```diff
@@ -631,6 +754,7 @@ def task_specific_observation(self, agent):
             joints_states = agent.get_internal_states()
```
Jialn, line 754: Remove the "else:" part
The only diff from "use full world states" is joints_states that contains angles and velocity of two wheels. Essentially another representation of position changes, velocity, yaw and yaw rate that are already in agent_vel. 
Sine we have the _use_full_states part, we should remove these. Option _use_full_states should also be removed as we dont need it anymore.

Jialn (changes requested):

pygazebo/pygazebo.cc
le-horizon: Was trying something due to error during closing env, not used any more, removed.

python/social_bot/envs/gazebo_base.py
le-horizon: somehow world.step + adding the camera causes random generator to be in a different state for the human rendering case vs rgb_array, even with the same random seed.

Now with this change and some other change in alf, pseudo randomness is exactly the same for the two modes.

python/social_bot/gazebo_agent.py
le-horizon: done.

python/social_bot/tasks.py
le-horizon: Since this code behavior has been in the code base for some time, I can't be sure that nobody is depending on this behavior.  I'm actually more comfortable leaving the else in, and adding this change in with flag protection.

python/social_bot/gazebo_agent.py
emailweixu, line 206: Agree with @Jialn that should keep GazeboBase clean. Specialized logic for particular task should be in other places.

python/social_bot/gazebo_agent.py
```diff
@@ -170,6 +190,25 @@ def get_observation(self, teacher, sentence_raw="hello"):
             obs = self.get_camera_observation()
         else:  # observation is pure low-dimentional states
             obs = teacher.get_task_specific_observation(self)
+            if self._goal_conditioned:
+                flat_obs = obs
+                obs = OrderedDict()
+                obs['observation'] = flat_obs
+                obs['achieved_goal'] = flat_obs[6:8]
```
emailweixu, line 197: Should have better way than `6:8`. Perhaps dictionary as you suggested.

python/social_bot/gazebo_agent.py
le-horizon: Good points.  Done all.

Thanks for the comments, and PTAL.

python/social_bot/gazebo_agent.py
le-horizon: Done.

python/social_bot/gazebo_agent.py
le-horizon: Done.

le-horizon (commented):
Hi Wei and JT, thanks for the detailed comments and nice suggestions.  Just addressed them.  PTAL.

python/social_bot/tasks.py
Jialn: It's fine to remove them, we don't need 2 similar implementation. Let's keep the code clean. 
 : )

python/social_bot/gazebo_agent.py
```diff
@@ -188,27 +180,15 @@ def get_observation(self, teacher, sentence_raw="hello"):
...
-                            (flat_obs[2:5], flat_obs[8:11], flat_obs[14:]),
-                            axis=0)
-                        obs['aux_achieved'] = np.concatenate(
-                            (flat_obs[0:2], np.expand_dims(flat_obs[5], 0),
-                             np.expand_dims(flat_obs[11], 0)),
-                            axis=0)
+        return obs
+
+    def _generate_goal_conditioned_obs(self, teacher):
+        tasks = teacher.get_task_groups()[0].get_tasks()
```
Jialn, line 190: do we need to assert here?

le-horizon: I moved the assert in get_observation_space, which is only called once per run, and is called before getting obs.

le-horizon (commented):
Addressed both.  (Removed the else in task_specific_observation.)

PTAL.

Thanks,
Le

python/social_bot/tasks.py
```diff
@@ -745,49 +745,41 @@ def task_specific_observation(self, agent):
                 pose = np.concatenate((pose, obj_pos), axis=0)
 
         agent_pose = np.array(agent.get_pose()).flatten()
-        if self._use_full_states or self._use_egocentric_states:
```
Jialn: option _use_full_states is not used now, so can be removed from GoalTask

le-horizon: Good point.  Done.

emailweixu (approved):

Jialn (approved):

PR #130, opened 2020-12-08
Update keybo_control

emailweixu:
Print control keys
Avoid keyboard conflict with matplotlib

### DIFF
--- a/python/social_bot/keybo_control.py
+++ b/python/social_bot/keybo_control.py
@@ -40,6 +40,19 @@ class KeyboardControl(PyKeyboardEvent):
         self._wheel_step = 0.5
         self._speed_decay = 0.9
         self._turning_decay = 0.6
+        self._done = False
+        logging.info("Control:" + """
+        W       : increase forward speed
+        S       : increase backward speed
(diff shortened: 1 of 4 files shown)


le-horizon (approved):
LGTM.  If @Jialn has feedback, we can probably incorporate after merge.

Thanks,
Le

Jialn:
Looks good to me : )

python/social_bot/keybo_control.py
```diff
@@ -40,6 +40,19 @@ def __init__(self):
...
+        W       : increase forward speed
+        S       : increase backward speed
+        A       : turn left
+        D       : turn right
+        E       : open/close gripper (if there is one)
+        R/F     : move robot arm joint (if there is one)
+        +       : increase the step size for W/S/A/D
+        -       + decrease the step size for W/S/A/D
+        Q       : quit
+        mouse   : gripper position (if there is one)
```
Jialn, line 54: This can be changed to "mouse: controls the yaw and pitch joints for robot arm (if there is one)", more precisly

PR #131, opened 2021-03-09
A simple parser based on regex

emailweixu:


le-horizon (dismissed):
LGTM.

On a high level, I think a deterministic logical form is probably better than natural language for both simple generation and simple parsing?
E.g. "left(agent, apple)" means apple is on the left of the agent.  "green(apple)" means apple is green.  "and(left(agent, apple), green(apple))" means a green apple is on the agent's left.

python/social_bot/parser.py
```diff
@@ -0,0 +1,125 @@
...
+
+    .. code-block:: python
+
+        parser = ReParser()
+        parser.add_word_class("OBJ", ["apple", "orange", "peach"])
+        parser.add_rule("an? <OBJ:OBJ_RIGHT> is on the right of an? <OBJ:OBJ_LEFT>", "OBJ_POS")
+        parser.add_rule("an? <OBJ:OBJ_LEFT> is on the left of an? <OBJ:OBJ_RIGHT>", "OBJ_POS")
+
+        result = parser.parse("a peach is on the right of an apple")
+        assert result.rule_name == "OBJ_POS")
```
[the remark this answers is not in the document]
emailweixu: removed

emailweixu:
> LGTM.
> 
> On a high level, I think a deterministic logical form is probably better than natural language for both simple generation and simple parsing?
> E.g. "left(agent, apple)" means apple is on the left of the agent. "green(apple)" means apple is green. "and(left(agent, apple), green(apple))" means a green apple is on the agent's left.

I guess we can always define the parsing rule to parse this kind of grammar if we need. But I don't feel it can make things much easier.

le-horizon (approved):
LG

python/social_bot/parser.py
```diff
@@ -0,0 +1,125 @@
...
+
+        Args:
+            word_class_name (str): name of the word class
+            words (list[str]): the words/phrases of this word class.
+        """
+        assert word_class_name not in self._word_classes, (
+            "duplicated world class name %s" % word_class_name)
+        self._word_classes[word_class_name] = words
+
+    def parse(self, sentence):
```
hnyu, line 103: probably should call .lower() on sentence?

emailweixu: Since all the words are from word IDs given the known vocabulary, they should not have wrong case.

hnyu (approved):

PR #132, opened 2021-03-10
Fix CI

emailweixu:
For some unknown reason, the CI failed recently (perhaps due to new version of Gazebo).
Switch to use alf docker image (which contains a fixed version of Gazebo) instead of installing new gazebo on the fly.


le-horizon (approved):
LG, thanks for fixing this, Wei!

PR #137, opened 2021-10-04
Upgrade gym: 0.15.3 -> 0.15.4

breakds:
# Motivation

To make it consistent with alf as alf is using gym 0.15.4.

# Solution

Bump up the version of gym in SocialRobots as well.

# Testing

Passed CI.


le-horizon (approved):
LG.

